PR #2, opened 2016-02-23
ENH: add Continuous Integration 

kmuehlbauer:
This PR adds CI to wradlib at github.


wradlib:
Since travis already starts python3 integration this fails, but python2.7 runs fine and the docs are built. So this is ready to merge.

Issue #8, opened 2016-02-24
Finalize python2/3 transition


kmuehlbauer:
Although travis runs fine with the current setup, there might be additional changes left.

- [x] make version aware, #9

kmuehlbauer:
@jjhelmus, you may dive in here, if you find time.

jjhelmus:
@kmuehlbauer Are these any Python2/3 issues in particular you are worried about.

kmuehlbauer:
For now everything runs fine on travis with the different python versions. So this can be closed. Any bugs that will show can be tackled in the future.

PR #9, opened 2016-02-24
ENH: make wradlib aware of its version

kmuehlbauer:
make wradlib aware of its version ,adapt setup.py, wradlib/__init__.py and doc/source/conf.py from pyart, add version and date to doc

This can go with issue  #8 , its somehow related to python2/3. Adapted setup.py with versioning applied.


wradlib:
#8

kmuehlbauer:
This needs to mbe merged to test deployment and versioning.

heistermann:
This PR involved to modify `__init__.py`, adding

`from .version import git_revision as __git_revision__`

When I try to import wradlib, this line throws an exception:

`ImportError: No module named version`

I don't have any idea about this module...is this some Python2/3 issue?

jjhelmus:
You need to trigger wradlib's build step which creates the `version.py` file.  Either use `python setup.py install` and changing directory so you do not pick up the source directory, or use `python setup.py build_ext --inplace` to do an in-place build.

kmuehlbauer:
@jjhelmus Thanks, I just started to write...

heistermann:
Thanks!

PR #10, opened 2016-02-24
FIX: fix references to bitbucket

kmuehlbauer:
All references to bitbucket are translated to github references


kmuehlbauer:
We should merge this asap, to have all documentation links point to the right location. If there is no response towards a further delay, I'll have this merged by tomorrow.

Issue #7, opened 2016-02-24
Add unittests for every submodule to increase coverage


kmuehlbauer:
Each submodule should have its corresponding unittest. This will increase coverage.

- [x] test_adjust
- [x] test_comp
- [x] test_dp
- [x] test_fill
- [x] test_verify
- [x] test_vpr
- [x] test_zr
- [x] add many test functions

kmuehlbauer:
Everything is done for the moment. More additions should be issued later on.

PR #11, opened 2016-02-25
ENH: Add test functions

kmuehlbauer:
This PR will be enhanced continuously...
resolves issue #7


kmuehlbauer:
All needed modules have their corresponding test-module. Many tests have been added, but many have to be added. Coverage in this PR is at 72.5 %.

PR #12, opened 2016-02-25
FIX: use anaconda channel within travis

kmuehlbauer:
several packages missing from https://repo.continuum.io/pkgs/free/linux-64/
switching to https://conda.anaconda.org/anaconda


wradlib:
It seems, that some dependencies (libnetcdf, setuptools, libgdal) are not correctly resolved using the free channel. By explicitely using the anaconda channel it can be resolved for the time being.

We should merge this in, to also have it also in the other PRs.

Issue #6, opened 2016-02-24
Finalize bitbucket/github transition


kmuehlbauer:
After moving wradlib from bitbucket to github there must be fixed all references to bitbucket to point to github. This will mostly relate to the docs. Also files should be removed, which are not needed any more in the github workflow.

- [x] fix references in doc, #10 
- [x] fix references in code/docstrings, #10 
- [x] testpypi deployment
- [x] revise deployment (docs)
- [x] fix links in docs, #22
- [x] add version switcher in docs, #22
- [x] check tagged builds
- [x] check main pypi

kmuehlbauer:
deployment works for testpypi-server, so this should also work for main pypi server. This has to be checked, before release of version 0.8.0

kmuehlbauer:
Deployment on tags works now.

kmuehlbauer:
Deployment of docs is working now.

kmuehlbauer:
Docs deployment works for untagged (devel) versions. We should add a tag (v0.7.6) to also check the tagged builds and upload to testpypi.

kmuehlbauer:
To test this, I'll set 0.7.6 as released in setup.py and tag this commit with v0.7.6.

kmuehlbauer:
Eureka, it works.

All implemented deployment functions are working as expected within travis:

- python setup.py install
- testrunner.py
- doc building und pushing
- building tagged versions
- (test)pypi upload

So, from the deployment side, we are ready to go!

PR #18, opened 2016-03-04
TST: further deployment testing

kmuehlbauer:
This PR adds a script for just installing wradlib within travis, and install package generating and uploading to pypi. All unittest and doc creation disabled.


kmuehlbauer:
#6

Issue #16, opened 2016-03-03
Check external links in Docs


kmuehlbauer:
At least the Radolan links are broken. Check and fix needed.

kmuehlbauer:
fixed by #22

kmuehlbauer:
This is merged with #22

PR #22, opened 2016-03-04
Finalizing deployment

kmuehlbauer:
This PR finalizes the deployment except the change to main pypi, we will have more test releases, so this will remain until before release of 0.8.0

With this PR also comes the revised links as mentioned in #16


kmuehlbauer:
OK, it seems that all problems are sorted out somehow. The doc is also not produced while in a PR.

kmuehlbauer:
The unittests in travis work again, if geos is missing in gdal the tests are overridden (not runned). See also #26.

To test doc building this pr has to be merged. I carefully tested doc-building locally. It should run fine, so I'll give it a try.

kmuehlbauer:
adds to #6

PR #24, opened 2016-03-07
Enable tests on Travis CI

jjhelmus:
The version of gdal on the OSGEO Anaconda.org channel comes with GEOS support.  One unit test was failing and was disabled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ install: source scripts/install.sh
 
 script:
     - export GDAL_DATA=/home/travis/miniconda2/envs/wradlib/share/gdal
-    #- xvfb-run coverage run --source wradlib testrunner.py -a
+    - xvfb-run coverage run --source wradlib testrunner.py -a
 
 after_success:
     - if [[ "$COVERALLS" == "true" ]]; then coveralls || echo "failed"; fi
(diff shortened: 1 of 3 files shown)


kmuehlbauer:
@jjhelmus 

Thanks for having a look.

I have checked travis output of my recent PR (#22) with this PR. Obviously there are used different versions of at least  gdal (1.11.2-py27_17 vs gdal-2.0.0-py2) and hdf4(hdf4-4.2.11-3 vs hdf4-4.2.11-0). 

For the error message (unittest.case.FunctionTestCase (ex_read_generic_netcdf) ... netcdf: 65536 is not a valid cdfid) I found this issue (Unidata/netcdf4-python#22) which indicates a shared library problem.

Do you happen to know, why conda did use the gdal 1.11.2 instead of gdal-2.0.0 in the other travis run? Should we pin the versions?

jjhelmus:
@kmuehlbauer I was having the hardest time yesterday getting the wradlib tests to pass either locally or on Travis CI but today using the package installs in your PR #22 everything just works.  Not sure if a new conda package was released or got updated or I was just missing something yesterday.  I don't see a need for this PR any more.  I'll close in preference to #22.

PR #28, opened 2016-03-09
Fixing some issues in zonalstats that appear to be Windows specific.

### LINKED ISSUE #25: Bug in ZonalDataBase when reading in layers from file using load_vector method
I think there is a bug in `ZonalDataBase.loadvector` (ll. 546 ff). Instead of

`ogr_copy_layer(ds_in, 0, self.src.ds)`
`ogr_copy_layer(ds_in, 1, self.trg.ds)`
`ogr_copy_layer(ds_in, 2, self.dst.ds)`

it should be

`ogr_copy_layer(ds_in, 0, self.dst.ds)`
`ogr_copy_layer(ds_in, 1, self.src.ds)`
`ogr_copy_layer(ds_in, 2, self.trg.ds)`

At least on my system... otherwise, you will have an error once you try to read `ZonalData` that has been dumped to a file before because the layers get mixed up. Can anyone reproduce this?

heistermann:
Without understanding what actually is behind this behaviour, this PR shold fix issue #25.

It also makes sure that the examples test whether loading ZonalData from file actually works. 

closes #25


wradlib/zonalstats.py
```diff
@@ -483,13 +483,22 @@ def _create_dst_datasource(self, **kwargs):
...
-
+        try:
+            tmp_trg_lyr.Intersection(src_lyr, self.tmp_lyr, options=['SKIP_FAILURES=YES',
+                                                                     'INPUT_PREFIX=trg_',
+                                                                     'METHOD_PREFIX=src_',
+                                                                     'PROMOTE_TO_MULTI=YES',
+                                                                     'PRETEST_CONTAINMENT=YES'],
+                                     callback=gdal.TermProgress)
+        except RuntimeError:
+            # Catch RuntimeError that was reported on gdal 1.11.1 on Windows systems
```
kmuehlbauer: OMG, I found the callback progress meter really nice. Doesn't work on windows?

kmuehlbauer:
I wonder, if we find more of the OS specific things.

Issue #39, opened 2016-04-05
Transition of example scripts to Jupyter Notebooks


kmuehlbauer:
New workflow that unifies examples, tutorials and teaching materials

- rewrite examples as jupyter notebooks, see #49

Tested workflow  #47 
- [x] convert/render notebooks with scripts/process_notebooks.sh
- [x] use scripts as testrunner input
- [x] rendered notebooks sphinxed with nbsphinx

Preparation:
- [x] WRADLIB_DATA set to examples/data at the moment
- [x] get_wradlib_data_file and get_wradlib_data_path function in ``util`` needed, to check local availability of data files/environment variable and ask whether data files should be downloaded from #38  
- [x] add new notebooks to sphinx' tocs
- [x] build docs

kmuehlbauer:
I just found https://github.com/spatialaudio/nbsphinx which might be worth testing.

The notebook would be rendered with `runipy` and the output written to doc/sources. There `nbsphinx` is doing the stuff off conversion.

Anyway, there are some things to consider:

- links within the notebook will work also in the sphinx html-output
- links relativ to other notebooks will work also in the sphinx-output
- function references (:func:`wradlib.georef.get_radolan_grid()`) if given in markdown will not be correctly transformed, we might need to use real links then. 
- If using nbformat v4 one could use rst within the cell and the rendered html will also work with :func:, but the notebook will look ugly.

While playing with `nbsphinx` it seems comfortable, also because we don't have to convert the resulting notebook (although we might convert it to V4). To cut a long story short, I'll adapt the workflow to use `nbsphinx`.

kmuehlbauer:
New Notebooks are to be announced here: #49

PR #47, opened 2016-04-08
Notebooks

kmuehlbauer:
This is a test PR with the proof of concept implementation of the notebook workflow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,12 @@ before_script:
     - flake8 --ignore=E501 wradlib
     - flake8 --ignore=E501 examples
     - flake8 --ignore=E501,E265 doc
+    # convert notebooks, render notebooks
+    - scripts/process_notebooks.sh
 
 script:
-    - export GDAL_DATA=/home/travis/miniconda2/envs/wradlib/share/gdal
+    - export GDAL_DATA=$HOME/miniconda2/envs/wradlib/share/gdal+
+    - export WRADLIB_DATA=$PWD/examples/data
     - xvfb-run coverage run --source wradlib testrunner.py -a
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,9 @@ matrix:
       env:
         - PYTHON_VERSION="3.5"
 
-install: source scripts/install.sh
+install:
+    - deactivate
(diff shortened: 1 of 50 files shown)


kmuehlbauer:
This is not going to work that easy.

It works only with python 3.5. With 2.7.11 and 3.4.2 it breaks with some weird numpy error (which I doubt is the real cause).
 
Locally python 2.7.11 and python 3.5.1 within conda works fine. I found similar issues all running on travis, but nothing what resolves this.

I also tried without conda-forge, to no avail. Also tampering with the PATH and other crazy things didn't have any effect. I'm lost somehow.

Next step is to setup a miminal "working/breaking" example in a clean repo.

kmuehlbauer:
Note to self: Just checked this without doing this within unittest. Locally no deprecation warnings are issued, when rendering the notebooks. So maybe we should first transform the notebooks to scripts and then run `testrunner` on that scripts. 

This also means to execute the notebooks-code twice:
1. the converted scripts in testrunner
2. second the notebooks in the renderer for the docs

We can later have a look, whats going wrong with unittests. Also we can bundle the scripts as examples.

kmuehlbauer:
This did it: ipython/ipython#8898

kmuehlbauer:
I put the workflow into issue #39.

@heistermann 
This works now side-by-side, examples and notebooks. So we are free to merge this (when the CI is happy) and add notebooks step by step, and removing the examples when the corresponding notebook works.

kmuehlbauer:
Another thing, what I need to add here: Should the notebooks be bundled with the source distro? Then we would need to add the notebooks/*.ipynb files to MANIFEST.in

kmuehlbauer:
@heistermann:

This is now ready from my side:
- the notebook workflow is shown by three notebooks. 
- WRADLIB_DATA used
- wradlib-data repo used
- use nbconvert and runipy for conversion/rendering of notebooks
- use nbsphinx (pandoc) for integration of rendered notebooks in docs
- use converted notebooks as testrunner input

As this is a add-on and examples and notebooks are working side-by-side there should be no problem by merging this. I'll pull this in by tomorrow morning.

heistermann:
Absolutely outstanding how you pulled this through... +1 from me!

Issue #63, opened 2016-07-07
Introduce wradlib error messages


kmuehlbauer:
We do many checks to be sure the given data applies to specific constraints.

For instance, we have the [`_check_polar_coords()`-function](https://github.com/wradlib/wradlib/blob/master/wradlib/georef.py#L705) inside `georef.py` module.

There we check several cpnstraints for azimuth and range arrays. We do also print out very specific error messages. Unfortunately we just `exit()` there, if something went wrong. 

We should raise a `wradlib error ` there. Then applications using this interface can react on those errors and conduct countermeasures. At the moment, one can only guess what went actually wrong.

Any suggestions or hints on the possible layout of the `wradlib error`?

kmuehlbauer:
OK, this seems not this big a problem, as I thought first.

`georef._check_polar_coords()` and `io.from_hdf5()` are the only functions which just use `exit()`. So we could just raise an appropriate error there.

We also have some wradlib errors already defined, eg. in [atten.py](https://github.com/wradlib/wradlib/blob/09b4caec7135175f9953cf0a34deaa92bcff4e7a/wradlib/atten.py#L33)

In the end, for the users it would be nice to know, if the error was raised within wradlib module or in some other python module. Yes, this is known from the stacktrace but would be good anyway.

kmuehlbauer:
solved by #64

PR #64, opened 2016-07-11
FIX: change exit()-calls to raise

kmuehlbauer:
Solves #63 in module `georef.py`

Changes the `exit()`-calls to `raise ValueError('some sophisticated ErrorString`). This enables the user to at least catch the error and handle it.


kmuehlbauer:
@heistermann Is this LGTY? It's working on my side.

heistermann:
+1 from me.

PR #66, opened 2016-07-11
ENH: copy notebook scripts to examples folder

kmuehlbauer:
add notebooks to distribution.

This adds notebooks and the converted python scripts to the `source distribution` (PyPi).


kmuehlbauer:
@ping heistermann

This works locally when creating a source disctribution (sdist), so it should also work in the PyPI deploy process.

heistermann:
Looks good to me - thanks!

PR #68, opened 2016-07-13
ENH: add reading capability of new radolan FZ product

kmuehlbauer:
This adds the FZ product to the radolan composit reader.

This is actually very preliminary. Means the two new header keys will probably be renamed in a future version, when their correct dwd-name is figured out.

This also adds a new grid specification (450 by 450), which wasn't defined before.

### DIFF
--- a/wradlib/georef.py
+++ b/wradlib/georef.py
@@ -1449,10 +1449,12 @@ def get_radolan_grid(nrows=None, ncols=None, trig=False, wgs84=False):
     """
 
     # setup default parameters in dicts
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
ping @heistermann Can merge this anyway? 

The coverage is just decreasing due to the two lines of added code, which isn't run now. We would need an example file. I'll try to get permission to use one from DWD.

heistermann:
Sorry, I missed this message - please merge at your discretion.

Issue #65, opened 2016-07-11
Remaining example scripts to be converted to notebooks


heistermann:
The following scripts from wradlib/examples still need to be converted to jupyter notebooks:

- [x] adjust_example
- [x] clutter_cloud_example
- [x] clutter_gabella_example
- [x] compose_by_height_example
- [x] coord
- [x] georef_example
- [x] gis_export_example
- [x] histo_cut_example
- [x] ipol_example
- [x] plot_ppi_example
- [x] verify_example
- [x] generic_netcdf_example
- [x] load_rainbow_example

Please check the box once you've converted one of these or if you are currently in the process of converting.

heistermann:
Resolved by Pull Request #69

PR #69, opened 2016-07-15
Translating remaining example scripts to jupyter notebooks.

heistermann:
This PR relates to issue #65.

Be aware that the conversion is not complete, yet. We should only merge this if issue #65 is completely resolved. However, I wanted to see how the PR behaves.


notebooks/wradlib_clutter_cloud_example.ipynb
```diff
@@ -0,0 +1,269 @@
...
+   "source": [
+    "import numpy as np\n",
+    "\n",
+    "import wradlib.vis as vis\n",
+    "import wradlib.clutter as cl\n",
+    "import wradlib.georef as georef\n",
+    "import wradlib.ipol as ipol\n",
+    "import wradlib.io as io\n",
+    "import wradlib.util as util\n",
+    "import matplotlib.pyplot as plt\n",
```
kmuehlbauer: breaking due to being plt instead of pl

kmuehlbauer:
@heistermann 
I just saw that we still pin hdf5 to 2.5.0 . This isn't necessary any more. It also keeps us from using the faster zonalstats with gdal 2.1.0. 

Could you please remove the pinning in the next iteration of this PR?

kmuehlbauer:
Nice, python 2.7 improved from 24 min to 20 min, python 3.4 gained a bit less, but python 3.5 took longer. 

I think this is somewhere within the notebooks running. I saw that mpl fontcache creation were running multiple times. Just keeping that for reference here.

Also the zonalstats polygon part is running almost twice as fast. Real improvement within gdal/geos.

wradlib/verify.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, r, az, sitecoords, proj, x, y, nnear=9):
         # compute the centroid coordinates in lat/lon
         bin_lon, bin_lat = georef.polar2centroids(r, az, sitecoords)
         # reproject the centroids to cartesian map coordinates
-        binx, biny, _ = georef.reproject(bin_lon, bin_lat,
+        binx, biny = georef.reproject(bin_lon, bin_lat,
                                          projection_target=proj)
```
kmuehlbauer: Flake8 issue is here, the indentation needs re-alignment.

wradlib/verify.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, r, az, sitecoords, proj, x, y, nnear=9):
         # compute the centroid coordinates in lat/lon
         bin_lon, bin_lat = georef.polar2centroids(r, az, sitecoords)
         # reproject the centroids to cartesian map coordinates
-        binx, biny = georef.reproject(bin_lon, bin_lat,
```
kmuehlbauer: Yeah, the trailing whitespace... :grinning:

kmuehlbauer:
> Coverage increased (+1.5%) to 79.438% when pulling 1b53cc3 on heistermann:master into 2916823 on wradlib:master.

Only 0.562 % away from the magic 80%.

kmuehlbauer:
@heistermann The notebooks are looking good. Merge, when you are ready.

PR #73, opened 2016-07-18
FIX: incorporate fz-product format description

kmuehlbauer:
DWD sent the format description of the FZ-product.

'VV' - "Vorhersagezeitpunkt" - "prediction time"
'MF' - "Modulflag" - "moduleflag

Also, the data is in RVP6 units like DX and RX. To convert from RVP6 to dBZ:

```python
dbZ = RVP6/2 - 32.5
```


kmuehlbauer:
@heistermann Everything runs smoothly, I'll pull this in.

Issue #70, opened 2016-07-15
Create tree structure for example notebooks and documentation?


heistermann:
There will be a lot of notebooks in the notebooks directory, and in the corresponding documentation. Should we think about organizing these in some kind of tree structure according to topics? Something like:

- First steps
- Georeferencing
- Quality control
- Visualisation
- ...

This way, users will have a better overview *where* they can or should look for what they need. We might also have a kind of course structure in the back of our minds when desiging a tree.

kmuehlbauer:
Seems very reasonable. :+1: 

The `radolan`-(sub)notebooks have already an higher-level document `radolan.ipynb` from which the sub-notebooks can be reached. This also goes nicely in the docs.

A tree structure would would definitely help users (and devs also). I suggest to tackle this when all notebooks are in place.

heistermann:
Very good idea - using "high-level" notebooks to define the directory tree for a specific topic. I also agree to address this once all notebooks are translated.

heistermann:
Maybe we can discuss a structure right here. 

These are my suggestions for high-level notebooks and the actual example notebooks that we currently have. Please note:
- some notebooks might fit in different topics and could therefore also be *linked* from different topics (e.g. wradlib_gis_export_example.ipynb might fit under RADOLAN and under File I/O),
- for those sections with a checkmark a high level notebook already exists,
- Some section only contain *one* notebook, yet - but additional notebooks might come in the future.
- Of course, additional sections might be created at any time if it makes sense.

- [ ] **Getting started**
- wradlib_introduction.ipynb
- wradlib_workflow.ipynb
- wradlib_get_rainfall.ipynb
- radolan_quickstart.ipynb

- [x] **RADOLAN: radolan.ipynb**
- radolan_format.ipynb
- radolan_grid.ipynb
- radolan_network.ipynb
- radolan_quickstart.ipynb
- radolan_showcase.ipynb
- wradlib_gis_export_example.ipynb
- simple_plot.ipynb

- [ ] **File I/O): file_io.ipynb**
- wradlib_radar_formats.ipynb
- wradlib_gis_export_example.ipynb
- wradlib_generic_netcdf_example.ipynb
- wradlib_load_rainbow_example.ipynb
- wradlib_reading_dx.ipynb

- [ ] **Georeferencing**
- wradlib_georef_example.ipynb
- wradlib_coords_example.ipynb

- [ ] **Visualisation**
- wradlib_plot_curvelinear_grids.ipynb
[...]

PR #75, opened 2016-07-21
Structure notebooks

kmuehlbauer:
As agreed on in #70, here is the first attempt.


kmuehlbauer:
Notebook tests needs fixing

kmuehlbauer:
@heistermann How do we want to handle the converted .py-scripts. Now they are created within the same folder as the original .ipynb. I suggest to leave them there and package them together with the .ipynb files for PyPI. We could remove the examples folder altogether then.

heistermann:
Agreed!

heistermann:
Merge at your discretion.

kmuehlbauer:
There are some flaws left, wrt to hyperlinks from notebooks to other docs (API etc.). I need to merge these changes here, if the CI is happy, to see which links need to be changed.

kmuehlbauer:
Something really weird happened. Somehow to example/data files were not deleted. But I couldn't remove them any more. Reverting seemed to make everything more cluttered. I did a hard reset to the commit before the breaking changes and will redo everything from clean slate.

PR #79, opened 2016-07-22
Notebook fixup 2

kmuehlbauer:
- fixing links in notebooks (inter-notebook and to docs)
- fixing links in docs
- making notebooks in docs pointing to tag version when tagged build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,12 +50,29 @@ after_success:
     - if [[ "$DOC_BUILD" == "true" ]]; then cd $TRAVIS_BUILD_DIR; scripts/build_docs.sh; fi
 
 deploy:
-  provider: pypi
-  server: https://pypi.python.org/pypi
-  user: wradlib
-  password:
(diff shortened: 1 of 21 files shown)


kmuehlbauer:
Here we go:

- For the notebooks in the wradlib-repo all inter-notebook-links are relative. This makes the links work within the notebook folder structure and also when rendering the docs.
- All links to sphinx-generated html-files are pointing to eg. http://wradlib.org/wradlib-docs/latest/generated/wradlib.io.read_generic_netcdf.html
- For tagged builds, the `latest` will be replaced by the tag number within the CI. This retains correct linkage for the different versions.

kmuehlbauer:
Green lights. 

@heistermann I really like to make a pre-release as suggested here #80. If your are happy with that, I'll go for it (self merges included).

kmuehlbauer:
I think I found a solution form the pre-releases. I added a second deploy-provider to .travis.yml.

The all merges into `wradlib/wradlib` will be deployed to https://testpypi.python.org/pypi/wradlib/. So we can check if a version is deployed correctly and users can more easily install devel versions.

PR #87, opened 2016-07-27
ENH: render notebooks inside unittests

kmuehlbauer:
This changes the way the notebooks are rendered. We do this now as part of the unittest-process. This safes us from doing the rendering part twice. Should result in a significantly reduced build time on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,10 +33,9 @@ install:
 
 before_script:
     - flake8 .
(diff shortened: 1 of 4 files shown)


kmuehlbauer:
Yes, build times significantly reduced. But it doesn't work that way. While running the notebooks the code coverage cannot be measured. So we are still bound to run the converted scripts to measure code coverage.

Issue #86, opened 2016-07-26
Ipol breaking due to changes in scipy 0.18.0


kmuehlbauer:
When using scipy 0.18.0 the ipol notebook breaks while handling nan-data. Needs reworking. Pinned scipy to 0.17.1 within travis in #84.

kmuehlbauer:
See this [travis-log](https://travis-ci.org/wradlib/wradlib/jobs/147456764).

kmuehlbauer:
@heistermann Did you had the chance to look into this? Any ideas?

heistermann:
@kmuehlbauer: Not yet, but asap. Thanks for pointing this out.

heistermann:
Apparently related to this scipy/scipy#6515

Problem occurs in case the parameter `nnearest` is larger than the number of source points. I'll try to find a more generic solution.

kmuehlbauer:
There were some changes in ckdtree implementation, but I‘m not sure, if this is effecting this. Let‘s see what the scipy devs say.

heistermann:
Addressed by Pull Request #95.

PR #95, opened 2016-08-24
FIX: Addresses issue 86 introduces by scipy 0_18_0.

heistermann:
Issue #86 was apparently introduced by a change in behaviour of cKDTree (see also scipy/scipy#6515).

Maximum number of neigbours (nnearest) is now set to maximum number of
source points during runtime. A warning is released in this case.

This change affects ipol.Idw, ipol.OrdinaryKriging and ipol.ExternalDriftKriging.

Akso introduced exceptions in case sources or targets are missing.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -18,7 +18,7 @@ source activate wradlib
 conda config --add channels conda-forge
 
 # Install wradlib dependencies
-conda install --yes gdal numpy scipy=0.17.1 matplotlib netcdf4 h5py
+conda install --yes gdal numpy scipy matplotlib netcdf4 h5py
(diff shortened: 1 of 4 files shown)


kmuehlbauer:
@heistermann Nice solution The coverage would slightly increase (or stay the same) when we introduce a little unit test.

heistermann:
I'll leave the remaining 0.07 % for tomorrow...

heistermann:
Well, that was a painful lesson. But I think it's ready to be merged - @kmuehlbauer: what do you think?

kmuehlbauer:
@heistermann Yes, it sometimes is painful. On the other side we gain a really clean and stable code base. 

Finally,  we can use github's squash merge.

PR #102, opened 2016-09-01
FIX: use OSR  IsSame function to evaluate SRS equality

kmuehlbauer:
The old approach using `AutoIdentifyEPSG` and `GetAuthorityCode` doesn't work for some reason. Using OGR `IsSame` works fine so far.


heistermann:
Looks fine to me!

kmuehlbauer:
OK, I'll check if the testpypi stuff works and bump the build number to 0.9.1

kmuehlbauer:
Worked nice: https://testpypi.python.org/pypi/wradlib

PR #106, opened 2016-09-22
Fixes two issues with older gdal and matplotlib versions.

heistermann:
Fixes an issue in wradlib_overlay.ipynb: the Albers projection is not available in gdal 1.10 which we are still using on the Open VM. Used Web Mercator instead.

Fixes an issue in wradlib_zonalstats_quickintro.ipynb: `magma` colormap is only availabel in the latest matplotlib releases. Used `gist_heat` instead.

Minor: Revised the tutorial on how to use jupyter notebooks with conda environments.

### DIFF
--- a/doc/source/jupyter.rst
+++ b/doc/source/jupyter.rst
@@ -24,8 +24,13 @@ Once these conditions are met, you can open a notebook: Launch a console or bash
 
 	> jupyter notebook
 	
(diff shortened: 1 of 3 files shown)


kmuehlbauer:
@heistermann Just one minor note on gdal 1.10. Please take into account, that ZonalStats will be painfully slow if you use many catchments. So reducing the number of processed catchments will speed up things a bit.

And, merge at your discretion.

PR #107, opened 2016-09-22
MIN: add presentation with earth curvature

kmuehlbauer:


kmuehlbauer:
@heistermann I added some markdown with the heading **Visualize Beam Propagation showing earth curvature**. Also stripped the two PPI's and BBF to only show the curvature stuff. I'll merge when the CI is happy.

PR #109, opened 2016-09-23
ENH: add reading groups to netcdf reader

kmuehlbauer:
This adds the ability to read the new `groups` feature in netcdf4 to the generic netcdf reader.

- introduced new function `read_netcdf_group` which has all the code of the former reader function
- inside `read_netcdf_group` is called in recursive manner to gather all nested groups.
- `read_generic_netcdf` is just a wrapper function around `read_netcdf_group` for opening and closing the file
- also fixed a minor issue with empty strings


kmuehlbauer:
@heistermann  OK, I'll merge when the new CI run is ready.

PR #114, opened 2016-10-24
Small change so that read_RADOLAN_composite also accepts a file handle

cchwala:
Not sure if this could be useful for others.

But since I want to work with the different archived RADOLAN files more easily (current `bin.gz` files and historic `tar.gz` files) I added the possibility to pass a file handle to `read_RADOLAN_composite`. This makes it possible to untar and unzip the historic RADOLAN files on the fly and avoids storing uncompressed data or creating temporary files from the tar.gz files. I tried the approach with the temporary files, but this small change to `read_RADOLAN_composite` is much more elegant.

I can however understand that this is just a edge-case and of no use to most other users so that a merge does not make sense.

### DIFF
--- a/wradlib/io.py
+++ b/wradlib/io.py
@@ -796,7 +796,7 @@ def read_radolan_header(fid):
     return header
 
 
-def read_RADOLAN_composite(fname, missing=-9999, loaddata=True):
+def read_RADOLAN_composite(f, missing=-9999, loaddata=True):
     """Read quantitative radar composite format of the German Weather Service
 
     The quantitative composite format of the DWD (German Weather Service) was
@@ -820,8 +820,8 @@ def read_RADOLAN_composite(fname, missing=-9999, loaddata=True):
 
     Parameters
     ----------
-    fname : string
-        path to the composite file
+    f : string or file handle
+        path to the composite file or file handle
     missing : int
         value assigned to no-data cells
     loaddata : bool
@@ -846,7 +846,9 @@ def read_RADOLAN_composite(fname, missing=-9999, loaddata=True):
     NODATA = missing
     mask = 0xFFF  # max value integer
 
-    f = get_radolan_filehandle(fname)
+    # If a file name is supplied, get a file handle
+    if type(f) == str:
+        f = get_radolan_filehandle(f)
 
     header = read_radolan_header(f)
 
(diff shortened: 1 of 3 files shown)


wradlib/io.py
```diff
@@ -846,7 +846,9 @@ def read_RADOLAN_composite(fname, missing=-9999, loaddata=True):
     NODATA = missing
     mask = 0xFFF  # max value integer
 
-    f = get_radolan_filehandle(fname)
```
kmuehlbauer: We can try to prevent the explicit type-check like this:

``` python
try:
    header = read_radolan_header(f)
except AttributeError:
    f = get_radolan_filehandle(f)
    header = read_radolan_header(f)
```

kmuehlbauer:
@heistermann Can we handle it this way? This won't break the API, but extend it.

wradlib/io.py
cchwala: Okay. Should I update my pull request, or will you add your suggested solution?

kmuehlbauer:
@cchwala

Yes, please update your PR. And thanks for this useful addition. Now I do not have to extract the monthly files anymore...

cchwala:
On my system, and also on travis-ci `test_radolan_coords` fails. I do not see how this is related, though.

heistermann:
I agree - I also don't see how your PR would break the test. Appears to be some accuracy issue... a quick fix would be to replace the `assertEqual` by `assertAlmostEqual` in the failing `test_radolan_coords`. Any thoughts about this, @kmuehlbauer ?

kmuehlbauer:
> Any thoughts about this, @kmuehlbauer ?

See #115

heistermann:
@cchwala Could you include this change into your PR so we can pull it in?

kmuehlbauer:
@cchwala @heistermann 

The [test](https://github.com/wradlib/wradlib/blob/master/wradlib/tests/test_georef.py#L368) should look something like this:

``` python
x, y = georef.get_radolan_coords(7.0, 53.0)
self.assertAlmostEqual(x, -208.15159184860158)
self.assertAlmostEqual(y, -3971.7689758313813)
x, y = georef.get_radolan_coords(7.0, 53.0, trig=True)
self.assertEqual(x, -208.15159184860175)
self.assertEqual(y, -3971.7689758313832)
```

Maybe a push with these additions will also retrigger the CI. Not sure why travis-ci is still waiting for status. Maybe some github-api issue.

kmuehlbauer:
@cchwala 

Thanks for adding this and working through the precision issue.

PR #118, opened 2016-11-08
Added default parameters for WX data format

franklinvv:
The DWD also supplies WX files with dimensions of 1100x900. When compared to the RX grid, the WX grid is displaced 80km towards the east, and expanded 100km northwards and southwards.


franklinvv:
Failed to adjust the documentation accordingly. Will provide a new pull request.

PR #124, opened 2016-11-15
FIX: Correct error in calculation of pulse volume.

heistermann:
We need to take half of the beam width in order to compute the base radius of our cone.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -79,6 +79,7 @@ def git_version():
 
     return GIT_REVISION
 
+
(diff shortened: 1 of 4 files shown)


kmuehlbauer:
@heistermann some `flake8` errors, but only one induced by this change. I can't reproduce this locally.

kmuehlbauer:
@heistermann One last trailing whitespace ... :grinning: :

PR #129, opened 2017-02-20
WIP: gpm_to_wrl

kmuehlbauer:
Do not merge yet!

This is tracking the changes in this feature branch.

We can further PR changes to  `gpm_to_wrl`-branch until we think it's ready. Then we can finally merge.

ToDo:

- [x] finalize (a mere clean-up) GPM-Notebook
- [x] add untitests for new functions
- [x] final docs-review
- [x] for the GPM notebook, we need to decide whether we want to keep the "intermediary" data layer implemented via `io_func.py`, or whether we want to directly manage the data as numpy arrays in the notebook.
- [x] ...


heistermann:
Thanks, @kmuehlbauer - I just started to review the changes you have applied so far. Sorry for the delay, I had to finalize a paper first...

heistermann:
I added another item to the checklist above. It is about whether we'd like to keep `io_func.py` to conveniently manage the data in the notebook. I'd yes, for the time being. We can replace it later if necessary.

kmuehlbauer:
@heistermann l'm would keep it also, but we should think about generalizing the work-flow for different gr types and trmm/gpm. Also encapsulating the final matching function would go in the same direction. Atm we have all utility functions pulled into wradlib and the matching done in the notebook.

kmuehlbauer:
@heistermann the non-conformal re-formatting of the arrays in `trafo.py` breaks the flake8-linter. If we want to stick to this layout, we need to prepend each line with `  # noqa: E201`. Also there are some under/over-idents which need to be treated (E127/E128).

kmuehlbauer:
> we need to prepend each line with # noqa: E201.

Append, I mean append...

kmuehlbauer:
@heistermann I'm much inclined to merge this into master, when the CI comes back happy.

Here is what is done so far:
- added unittests for the introduced functions, 
- reviewed docs and docstrings in the changed modules
- made changes to the CI-builds, introduced python3.6 to matrix
- many minor fixes which are needed in master to step further with other PRs

I think the notebook can mature also in master. WDYT?

PR #130, opened 2017-03-02
ENH: add contour, contourf, cg options to `plot_ppi` and `plot_rhi`

kmuehlbauer:
*UPDATE:*

I merged the cg_plot functions into `plot_ppi` and `plot_rhi` functions, available via `cg=True` keyword. Also added `contour` and `contourf` to `plot_ppi` and `plot_rhi`. Grid/Subplot also available now.

Adapted Notebooks and Tests.

Deprecated `plot_cg_ppi` and `plot_cg_rhi`. Deleted my now obsolete comments.


kmuehlbauer:
@heistermann 

While thinking about this and looking at our visualisation functions `plot_ppi` and `plot_rhi` as well as their counterparts `plot_cg_ppi` and `plot_cg_rhi`it seems more than reasonable to me to reduce this to only two functions named `plot_ppi`and `plot_rhi`.

My suggestion is to implement the three different plotting strategies (`pcolormesh`, `contour` and `contourf`) within these two remaining functions as already laid in the proposed change of `plot_cg_rhi`.  This accounts as well to the new `overplot` keyword and to the strategy to plot on specific subplots.

The plotting on curvelinear grid would then be triggered via the additional keyword *cgplot=True*. 

Pros:
- `plot_ppi` and `plot_rhi` remain the one and only plotting functions of polar radar data.
- `plot_ppi` and `plot_rhi`will be backwards compatible, means we only add to the API.
- reduce of code duplication within the deprecated `plot_cg_ppi` and `plot_cg_rhi`.

Cons:
- another bunch of keywords
- different returns (ax, pm) vs. (cgax, caax, paax, pm) which might be confusing

WDYT?

kmuehlbauer:
The `plot_rhi` function after applying all necessary changes:

```python
def plot_rhi(data, r=None, th=None, th_res=None, autoext=True, refrac=True,
             rf=1., fig=None, ax=111, func='pcolormesh', cgplot=False,
             **kwargs):
    """Plots a Range Height Indicator (RHI).

    The implementation of this plot routine is in cartesian axes and does all
    coordinate transforms beforehand. This allows zooming into the data as well
    as making it easier to plot additional data (like gauge locations) without
    having to convert them to the radar's polar coordinate system.

    `**kwargs` may be used to try to influence the
    :func:`matplotlib.pyplot.pcolormesh` routine under the hood.

    Parameters
    ----------
    data : np.array
        The data to be plotted. It is assumed that the first dimension is over
        the elevation angles, while the second dimension is over the range bins
    r : np.array
        The ranges. Units may be chosen arbitrarily. If None, a default is
        calculated from the dimensions of `data`.
    rf: float
        if present, factor for scaling range axis
        defaults to 1.
    th : np.array
        The elevation angles in degrees. If None, a default is
        calculated from the dimensions of `data`.
    th_res : float or np.array of same shape as `th`
        In RHI's it happens that the elevation angles are spaced wider than
        the beam width. If this beam width (in degrees) is given in `th_res`,
        plot_rhi will plot the beams accordingly. Otherwise the behavior of
        :func:`matplotlib.pyplot.pcolormesh` assumes all beams to be adjacent
        to each other, which might lead to unexpected results.
    autoext : True | False
        This routine uses :func:`matplotlib.pyplot.pcolormesh` to draw
        the bins.
        As this function needs one set of coordinates more than would usually
        provided by `r` and `az`, setting ´autoext´ to True automatically
        extends r and az so that all of `data` will be plotted.
    refrac : True | False
        If True, the effect of refractivity of the earth's atmosphere on the
        beam propagation will be taken into account. If False, simple
        trigonometry will be used to calculate beam propagation.
        Functionality for this will be provided by functions
        :meth:`wradlib.georef.arc_distance_n` and
        :meth:`wradlib.georef.beam_height_n`, which assume distances to be
        given in meters. Therefore, if `refrac` is True, `r` must be given
        in meters.
    fig : matplotlib Figure object
        If given, the RHI will be plotted into this figure object. Axes are
        created as needed. If None, a new figure object will be created or
        current figure will be used, depending on "ax".
    ax : matplotlib Axes object | matplotlib grid definition
        If matplotlib Axes object is given, the RHI will be plotted into this
        axes object.
        If matplotlib grid definition is given (nrows/ncols/plotnumber),
        axis are created in the specified place.
        Defaults to '111', only one subplot/axis.
    func : str
        Name of plotting function to be used under the hood.
        Defaults to 'pcolormesh'. 'contour' and 'contourf' can be selected too.
    cgplot : True | False
        If True, the data will be plotted on curvelinear axes.

    See also
    --------
    create_cg : creation of curvelinear grid axes objects

    Returns
    -------
    ax : matplotlib Axes object
        The axes object into which the RHI was plotted, if cgplot is False.
    cgax : matplotlib Axes object
        curvelinear Axes (r-theta-grid), if cgplot is True.
    caax : matplotlib Axes object
        Cartesian Axes (x-y-grid) for plotting cartesian data,
        if cgplot is True.
    paax : matplotlib Axes object
        The parasite axes object for plotting polar data, if cgplot is True.
        All data in polar format must be plottet to this axis
    pm : matplotlib QuadMesh object
        The result of the pcolormesh operation. Necessary, if you want to
        add a colorbar to the plot.

    Examples
    --------
    See :ref:`notebooks/visualisation/wradlib_plot_curvelinear_grids.ipynb`.
    """
    # autogenerate axis dimensions
    if r is None:
        d1 = np.arange(data.shape[1], dtype=np.float)
    else:
        d1 = np.asanyarray(r.copy())

    if th is None:
        # assume, data is evenly spaced between 0 and 90 degree
        d2 = np.linspace(0., 90., num=data.shape[0], endpoint=True)
        #d2 = np.arange(data.shape[0], dtype=np.float)
    else:
        d2 = np.asanyarray(th.copy())

    if autoext & ('pcolormesh' in func):
        # extend the range by the delta of the two last bins
        x = np.append(d1, d1[-1] + d1[-1] - d1[-2])
        # RHIs usually aren't cyclic, so we best guess a regular extension
        # here as well
        y = np.append(d2, d2[-1] + d2[-1] - d2[-2])
    else:
        # hopefully, the user supplied everything correctly...
        x = d1
        y = d2

    if th_res is not None:
        # we are given a beam resolution and thus may not just glue each
        # beam to its neighbor
        # solving this still with the efficient pcolormesh but interlacing
        # the data with masked values, simulating the gap between beams
        # make a temporary data array with one dimension twice the size of
        # the original
        img = np.ma.empty((data.shape[0], data.shape[1] * 2))
        # mask everything
        img.mask = np.ma.masked
        # set the data in the first half of the temporary array
        # this automatically unsets the mask
        img[:, :data.shape[1]] = data
        # reshape so that data and masked lines interlace each other
        img = img.reshape((-1, data.shape[1]))
        # produce lower and upper y coordinates for the actual data
        yl = d2 - th_res * 0.5
        yu = d2 + th_res * 0.5
        # glue them together to achieve the proper dimensions for the
        # interlaced array
        y = np.concatenate([yl[None, :], yu[None, :]], axis=0).T.ravel()
    else:
        img = data

    # fix reference for contour functions
    if 'contour' in func:
        x += (x[1] - x[0]) / 2
        y += (y[1] - y[0]) / 2

    # axes object given
    if isinstance(ax, mpl.axes.Axes):
        if cgplot:
            caax = ax.parasites[0]
            paax = ax.parasites[1]
    else:
        if fig is None:
            # create new figure if there is only one subplot
            if ax is 111:
                fig = pl.figure()
            else:
                fig = pl.gcf()
        if cgplot:
            # create curvelinear axes
            ax, caax, paax = create_cg('RHI', fig, ax)

            # this is in fact the outermost thick "ring" aka max_range
            ax.axis["lon"] = ax.new_floating_axis(1, np.max(x) / rf)
            ax.axis["lon"].major_ticklabels.set_visible(False)
            # and also set tickmarklength to zero for better presentation
            ax.axis["lon"].major_ticks.set_ticksize(0)
        else:
            ax = fig.add_subplot(ax)

    # coordinates for all vertices
    xx, yy = np.meshgrid(x, y)

    plax = ax
    if refrac:
        # observing air refractivity, so ground distances and beam height
        # must be calculated specially
        xxx = georef.arc_distance_n(xx, yy) / rf
        yyy = georef.beam_height_n(xx, yy) / rf
        if cgplot:
            plax = caax
    else:
        if cgplot:
            xxx, yyy = np.meshgrid(y, x)
            yyy /= rf
            img = img.transpose()
            plax = paax
        else:
            # otherwise plane trigonometry will do
            xxx = xx * np.cos(np.radians(yy)) / rf
            yyy = xx * np.sin(np.radians(yy)) / rf

    # plot the stuff
    plotfunc = getattr(plax, func)
    pm = plotfunc(xxx, yyy, img, **kwargs)

    # return references to important and eventually new objects
    if cgplot:
        # set bounds to maximum
        ax.set_ylim(0, np.max(x) / rf)
        ax.set_xlim(0, np.max(x) / rf)

        # show curvelinear and cartesian grids
        # makes no sense not to plot, if we made such a fuss to get that handled
        ax.grid(True)
        caax.grid(True)
        return ax, caax, paax, pm
    else:
        return ax, pm

```

kmuehlbauer:
@heistermann This is now working smoothless. I'll add some unittests for the deprecated functions later today. I'll approve this when I think I'm done.

kmuehlbauer:
@heistermann Nice, you can't approve your own PRs.

PR #139, opened 2017-03-27
WIP: testrunner fixup

kmuehlbauer:
Testing how to make single-notebook-test testsuites running in separate processes with Xvfb on Travis-Ci.

1. With this change, every `unittest.TestSuite` is started in a separate subprocess. 

2. But for the memory consuming notebooks (because of the sheer amount of notebooks) this also breaks the memory limit of travis-ci (all notebooks are run in the same `unittest.TestSuite` / subprocess). So every notebook is started as single `unittest.TestSuite` in a separate process.

3. To get the headless mode with `xvfb` in the subprocesses, we need to export the `DISPLAY` and start `xvfb` before starting the tests.

4. To get `coverage` working with subprocesses, we need to state this in the `.coveragerc`. 

5. To get the correct `exit_status` from `run_test.sh`, we need to retrieve it from the `coverage`-calls.

This is working now, finally!

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,6 @@
 [run]
+concurrency = multiprocessing
+parallel = true
 source = wradlib
 omit =
     */setup.py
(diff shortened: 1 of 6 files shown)


kmuehlbauer:
@heistermann 

Finally...No code changes, just maintenance in testing/code-coverage. Please have a quick look. I'll merge tomorrow, if there are no objections.

PR #137, opened 2017-03-22
ENH: Raster handling

kmuehlbauer:
This PR handles the issues outline in #135 

new functions
- [x] `georef.create_raster_dataset` (data, coords, proj -> dataset)
- [x] `georef.extract_raster_dataset` (dataset -> data, coords, proj)
- [x] `georef.set_raster_origin` (flip data and coords, if necessary)
- [x] `georef.reproject_raster_dataset` (and resampling)
- [x] `io.write_raster_dataset` (write to disc)
- [x] `util.get_raster_origin` (returns `upper` or `lower`)
- [x] add unit-tests

deprecated functions
- [x] deprecate `io.to_AAIGrid` and `to_GeoTIFF`
- [x] deprecate `io.read_raster_dataset` (the magic needed in this function isn't worth the work)
- [x] deprecate `georef.resample_raster_dataset`

other fixes
- [x] fix `zonalstats.DataSource.dump_raster` to use new functionality
- [x] fixed `io.read_safnwc` (proj4/wkt-handling)
- [x] update `notebooks/fileio/wradlib_gis_export_example.ipynb`
- [x] update `georef.read_gdal_values` to read multi band data
- [x] fix `util.find_bbox_indices` to account for `lower` or `upper` origin data/coords
- [x] fix `ipol.cart2irregular_spline` to account for data/coords origin

PR #140, opened 2017-03-30
ENH: added option to read rainbow files from file handle

cchwala:
As discussed, [here](https://github.com/cchwala/wradlib/commit/d55678fda5afc9945844f258586d6a47da68e4f3#commitcomment-21557664), this is now the PR to make it possible to read Rainbow files from a file handle, e.g. when reading from a zipped archive.

I assumed that the low level functions do not need the `apichange_kwarg` since nobody ever uses them directly and they only get called by `read_Rainbow`. It would be hard (or impossible without writing to a temp file) to transform the `fid` file handles into the now deprecated `filename` strings because in many cases (or at least in my application) there is no path to the files which e.g. may belong to a zipped archive.

After rebasing on master I cannot run the tests. They fail with "The process has forked and you cannot use this CoreFoundation functionality safely. You MUST exec().". Hence, I hope that the CI runs will complete.

### DIFF
--- a/wradlib/io.py
+++ b/wradlib/io.py
@@ -1542,15 +1542,15 @@ def get_RB_blob_from_string(datastring, blobdict):
     return data
 
 
-def get_RB_blob_from_file(fid, blobdict):
+def get_RB_blob_from_file(f, blobdict):
     """
     Read BLOB data from file and return it with correct
     dataWidth and shape
 
     Parameters
     ----------
-    fid : file handle
-        File handle of Data File
+    f : string or file handle
+        File handle of or path to Rainbow file
     blobdict : dict
         Blob Dict
 
@@ -1560,10 +1560,21 @@ def get_RB_blob_from_file(fid, blobdict):
         Content of blob as numpy array
     """
 
+    # Try to read the data from a file handle
     try:
+        f.seek(0, 0)
+        fid = f
         datastring = fid.read()
-    except:
-        raise IOError('Could not read from file handle')
+    except AttributeError:
+        # If we did not get a file handle, assume that we got a filename,
+        # get a file handle and read the data
+        try:
+            fid = open(f, "rb")
+            datastring = fid.read()
+            fid.close()
+        except IOError:
+            print("WRADLIB: Error opening Rainbow file ", f)
+            raise IOError
 
     data = get_RB_blob_from_string(datastring, blobdict)
 
@@ -1653,7 +1664,6 @@ def get_RB_header(fid):
     return xmltodict.parse(header)
 
 
-@util.apichange_kwarg('0.9.2', par='filename', typ=str, expar='f')
 def read_Rainbow(f, loaddata=True):
     """Reads Rainbow files files according to their structure
 
@@ -1684,6 +1694,9 @@ def read_Rainbow(f, loaddata=True):
     --------
 
     See :ref:`notebooks/fileio/wradlib_load_rainbow_example.ipynb`.
+
+    .. versionchanged 0.10.0
+       Added reading from file handles.
 
     """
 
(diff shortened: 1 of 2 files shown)


cchwala:
Sorry for the iterative fixing based on CI errors. As stated above, test do currently not run locally on my machine. Feel free to squash the commits. The first one is labled wrongly anyway. "WIP" should now maybe be "ENH", when it's finally working...

kmuehlbauer:
> After rebasing on master I cannot run the tests. They fail with "The process has forked and you cannot use this CoreFoundation functionality safely. You MUST exec().". Hence, I hope that the CI runs will complete.

@cchwala Are you on OSX? On my linux box everything runs fine, when I start the tests on the console with ```python testrunner.py -a```. 
The changes were introduced with #139. We were forced to run testsuites in subprocesses to overcome the overall memory limit of 3GB on travis-ci.

cchwala:
@kmuehlbauer Yes OSX. I will try on a Linux machine. Should I open an issue concerning the tests on OSX?

kmuehlbauer:
@cchwala This is a problem obviously with OSX (and maybe with windows). I found this: http://stackoverflow.com/questions/16254191/python-rpy2-and-matplotlib-conflict-when-using-multiprocessing

But you can try `xvfb` to make the tests headless. See `.travis.yml`. 

You can open an issue, and we can reference everything there.

cchwala:
@kmuehlbauer Sorry, I am very busy this week. Would it suffice to finish this PR next monday?

kmuehlbauer:
@cchwala I would really like to roll out wradlib release `0.10.0.` on next monday. I could pull your additions from your fork and add my suggested changes. Are you OK with this way of working?

cchwala:
@kmuehlbauer Okay, I will do it till Friday evening/night. It does not look like too much work anyway and since I started this I also want to bring it to an end ;-).

kmuehlbauer:
@cchwala This would fit. 

@heistermann Regarding the **apichange**. Since we do not have something for `*args` we do not have much options. From the workflow it seems unnecessary to open/close the file multiple times. So I tend to silently break the API for the low-level functions (leave the `get_RB_blob_from_file` handle both `filename` and `fid`). Adding `.. versionchanged:: 0.10.0` in all changed functions should suffice.

cchwala:
@kmuehlbauer #137 produces a diff for `io.py` that is the whole file content. I have a hard time to solve the merge conflict since its 2000 lines of code vs 2000 lines of code. Right now I am assuming that you only changed things concerning `io.write_raster_dataset` within `io.py` and will manually resolve the conflicts. Any additional info from your side?

cchwala:
@kmuehlbauer Merge done. Test should complete, at least they now do on my Mac. Now on to the requested changes.

wradlib/io.py
```diff
@@ -1616,53 +1610,51 @@ def get_RB_blobs_from_file(filename, rbdict):
...
+        raise IOError('Could not read from file handle')
 
     xmltodict = util.import_optional('xmltodict')
 
     return xmltodict.parse(header)
 
 
-def read_Rainbow(filename, loaddata=True):
-    """"Reads Rainbow files files according to their structure
+@util.apichange_kwarg('0.9.2', par='filename', typ=str, expar='f')
```
cchwala: Removed the decorator according to your request. But isn't the decorator also mean to catch the case where people have `read_Rainbow(filename='my_file.azi')` in their code?

kmuehlbauer: *filename* is no *kwarg*, so it isn't catched by this decorator. We would need some decorator for *args*.

kmuehlbauer:
@cchwala Sorry for the inconvenience with the io.py. I remember having issues with the line endings. That's probably the cause for the large difference.

kmuehlbauer:
@cchwala Thanks for your contribution! Hoping to see more from your side 👍.

PR #147, opened 2017-05-11
MAINT: replace `runipy` workflow by `execute`-API workflow

kmuehlbauer:
Since introduction of the [execute-API](http://nbconvert.readthedocs.io/en/latest/execute_api.html#example) it is possible to execute notebooks within unittests. 

This way we can test the notebooks and in the same time get them rendered.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ before_script:
     - flake8 .
     - export GDAL_DATA=$HOME/miniconda/envs/wradlib/share/gdal
     - export WRADLIB_DATA=$HOME/wradlib-data
-    - export WRADLIB_NOTEBOOKS=$HOME/notebooks-render
+    - export WRADLIB_NOTEBOOKS=$TRAVIS_BUILD_DIR/notebooks-render
     # duplicate notebooks folder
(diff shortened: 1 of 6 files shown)


kmuehlbauer:
@heistermann 

Coverage testing of notebooks is rather hard to achieve. The wrapping of `testrunner.py` with `coverage run` doesn't work for notebook-executions. Neither with `runipy` nor with `nbconvert --execute` or their API-versions.

But we would get a significant speed-up by using the  [execute-API](http://nbconvert.readthedocs.io/en/latest/execute_api.html#example) for rendering the notebooks and do code coverage testing in the same time.

To achieve this I implemented the code coverage starting/stopping within `testrunner.py`. To get the notebooks started I had to add one code-cell with the coverage startup-sequence at the top of each notebook before feeding it to the renderer. This cell is then removed before the write-process. For the other tests the code coverage is simply used with the coverage-API functions.

In `scripts/run_tests.sh` all `.coverage.*`-files are moved into one directory and then `coverage combine` is called. So we get the code coverage calculated into `.coverage`.

Especially the notebook-part seems a bit of a hack, but this works very well locally. Let's see how this works on the CI.

kmuehlbauer:
@heistermann Works! I'll merge this on monday, if there are no objections from your side.

PR #148, opened 2017-05-15
ENH: use wradlib docker image to run build process

kmuehlbauer:
Use docker image from [wradlib-docker](https://github.com/wradlib/wradlib-docker) to build and test wradlib on Travis CI.


kmuehlbauer:
@heistermann 

This is also working now. Although the codecov-checks are reported waiting, the codecov-bot reported the coverage correctly.

So, we would need to decide if we want to use the docker processing in the CI.

Pros:
- faster download (at least for the python3.6 version)
- repeatable builds with same environment
- runs faster (no explanation so far)
- can be used locally 
- ...

Cons:
- docker-repo needs to be maintained, docker-image also needs testing
- ...

WDYT?

heistermann:
Not sure that I understand everything. Using the wradlib-docker image to test and build wradlib sounds reasonable, particularly if we want to host a docker image anyway as one option to deploy wradlib.

However, I am still confused what is tested at which point and for which purpose... so this is how I understand it:

1. We create a wradlib-docker image (as for *now*, this does not involve testing). "Creating" implies installing the wradlib runtime environment, wradlib itself, and finally exporting the image.

2. In CI for wradlib, we would then just pull the docker-image instead of reinstalling the wradlib runtime environment each time Travis is called. However, after pulling the docker image, we need to re-install the current wradlib version that is to be actually tested (in `install_docker.sh`?). Is that right?

My question is how we make sure that the environment on the docker image is consistent with the wradlib code that is currently tested? Wouldn't we need to run all the tests when we build the docker image which would then, however, be a duplicate effort? But maybe I don't get it, yet. Anyway, I guess for answering the question, we need to make a decision on how much we would like to cut down the time that is needed to test a new commit.

kmuehlbauer:
> We create a wradlib-docker image (as for now, this does not involve testing). "Creating" implies installing the wradlib runtime environment, wradlib itself, and finally exporting the image.

Yes, I think we would need one dedicated docker-image for every supported python version. Then there would be tags for every minor wradlib-version. These docker images should be build and tested automatically. We could release updates to that images depending on updates of wradlib's dependencies and/or on a regular schedule. If we release a new wradlib version a new docker image is created.

> In CI for wradlib, we would then just pull the docker-image instead of reinstalling the wradlib runtime environment each time Travis is called. However, after pulling the docker image, we need to re-install the current wradlib version that is to be actually tested (in install_docker.sh?). Is that right?

This is what is intended. ATM there is only a py36 docker images and for py27 and py35 the runtime is reinstalled. But that won't be needed if we have images for every supported python version.

> My question is how we make sure that the environment on the docker image is consistent with the wradlib code that is currently tested? Wouldn't we need to run all the tests when we build the docker image which would then, however, be a duplicate effort?

Actually this is only done once, when creating the image. Let's say we create an image `wradlib-py36:0.10.0`. This image is tested with wradlib version 0.10.0. Then this image is used within the CI for all subsequent PRs/merges (for py36) until a new image is created.

Since we would use one image a rather long time, we avoid possible breaks due to changes in the runtime and get repeatable results.

heistermann:
Sounds good. So would you do the transformation to use docker images at first only for python 3.6, or would you change the procedure for all python versions at once?

kmuehlbauer:
It would be better to have it all in place before making the change, imho.

Then I'll make the necessary changes to  [wradlib-docker](https://github.com/wradlib/wradlib-docker). This will surely take some time to get the whole workflow up and running. I'll keep you posted.

kmuehlbauer:
Comparing the latest non-docker PR #163 (overall runtime 1h 31min 10sec) with the current docker implementation is quite encouraging. The runtime was onle 50min 30sec. It reduced the runtime almost by a factor of two. Using dedicated docker images (python 2.7, 3.5, 3.6) we can expect another speedup of several minutes.

kmuehlbauer:
So with the latest improvements in [wradlib/wradlib-docker](https://github.com/wradlib/wradlib-docker), we have one docker image for every python version (2.7, 3.5 and 3.6). With this implemented in the travis docker workflow, we get the processing time own to 45min total (which is only one half of the former used time). 

Every matrix build is now running between 14 to 16 minutes wich is quite a fantastic improvement to the former 25-35 minutes.

One thing what need to be resolved is when the docker images need a rebuild. From my point of view, we can stay with the manual update of the docker file.

The docker file needs update when:

- wradlib need an updated build/run dependency to work (eg. new feature)
- we consider a dependency upgrade as necessary (eg. bug-fix)
- found a bug with the docker image itself

kmuehlbauer:
I'm thinking about having a commit switch like travis' [ci skip](https://docs.travis-ci.com/user/customizing-the-build#Skipping-a-build), to decide if we want to use the docker image, or the normal travis/conda approach. Let's see, how we can achieve that.

PR #154, opened 2017-06-14
FIX: add 'QN' header token to RADOLAN reader

### LINKED ISSUE #153: Missing "QN" flag in wradlib.io.get_radolan_header_token()
There is an issue in wradlib.io which prevents RADOLAN and RADVOR files with set QN flag to be read.

This is easily fixed by adding the "QN" flag to head dictionary in `wradlib.io.get_radolan_header_token()`

Without this flag, the end points from other header parts are read wrong, resulting in ValueErrors. 

The flag also has to be added to `wradlib.io.parse_DWD_quant_composite_header(header)` to further work with it in the output data of `wradlib.io.read_RADOLAN_composite(f, missing=-9999, loaddata=True)`

kmuehlbauer:
Add 'QN' header token to RADOLAN reader

closes #153


kmuehlbauer:
@heistermann Could you have one last look? Then this can be merged away...

PR #155, opened 2017-06-19
WIP: Reorganize Submodules

kmuehlbauer:
This is a first attempt to reorganize wradlib's submodules.

As an extension to the suggested refactoring in https://github.com/wradlib/wradlib/pull/150#issue-233813141, we should also split already huge submodules into smaller parts. This would make code changes/reviews easier and wradlib better maintainable.

I started with the `wradlib.io` submodule first. Suggestions very welcome.

- [x] wradlib.georef (GR)
    - misc, polar, projection, raster, vector, rect, satellite
- [x] wradlib.io (IO)
    - gdal, hdf, misc, netcdf, radolan, rainbow

Update: submodules which do not need splitting have been removed.

Helper functions from `wradlib.zonalstats` could be moved to a more appropriate location.

yet undecided where or whether to move
- numpy_to_pathpatch (ZS) (for visualisation, so could be moved to `wradlib.vis`)

- mask_from_bbox (ZS) (returns rectangular mask, no exact solution but much faster than the below)
- get_clip_mask (ZS) (exact solution, but slower than the above)
- angle_between (ZS) (needed by mask_from_bbox)

- get_bbox (ZS)
- grid_centers_to_vertices  (ZS) (which does the same as `wradlib.georef.polar.centroid2polyvert`, only counterclockwise, so a unification would be reasonable)


kmuehlbauer:
@heistermann 

The reorganization of the io-module was successful so far. I think, that we should have the gdal/ogr read/write capabilities also within one new submodule. But I need to think about this a bit more.

So atm we have five new submodules below `wradlib.io`. All needed functions are available by import from `wradlib.io`. To get the other functions one would need to import from eg. `wradlib.io.radolan`. So we can maintain API compatibility while having separated by functionality. 

Please add your comments and suggestions also for splitting other modules.

kmuehlbauer:
I updated the first post. It turns out, that only the modules `wradlib.io` and `wradlib.georef` need submodules. The gdal/ogr functions from module `wradlib.zonalstats` can be moved to `wradlib.georef` and `wradlib.io`. There are also functions which can be unified.

kmuehlbauer:
Close in favour of #159.

PR #159, opened 2017-06-23
Reorganize modules

kmuehlbauer:
This PR takes over from #155

This is the second attempt to reorganize wradlib's submodules.

As an extension to the suggested refactoring in https://github.com/wradlib/wradlib/pull/150#issue-233813141, we should also split already huge submodules into smaller parts. This would make code changes/reviews easier and wradlib better maintainable.

Submodules created for:

- [x] wradlib.georef (GR)
    - misc, polar, projection, raster, vector, rect, satellite
- [x] wradlib.io (IO)
    - gdal, hdf, misc, netcdf, radolan, rainbow

yet undecided where or whether to move
- numpy_to_pathpatch (ZS) (for visualisation, so could be moved to `wradlib.vis`)

- mask_from_bbox (ZS) (returns rectangular mask, no exact solution but much faster than the below)
- get_clip_mask (ZS) (exact solution, but slower than the above)
- angle_between (ZS) (needed by mask_from_bbox)

- get_bbox (ZS)
- grid_centers_to_vertices  (ZS) (which does the same as `wradlib.georef.polar.centroid2polyvert`, only counterclockwise, so a unification would be reasonable)


kmuehlbauer:
Although this reorganizes functions below `wradlib.io`, `wradlib.georef`and `wradlib-zonalstats`it doesn't break the API. Everything is working nicely. Further improvements in subsequent PR's.

kmuehlbauer:
@heistermann I'll merge this as agreed in the last online dev-meeting.

PR #163, opened 2017-06-28
ENH: introduce flags to code coverage

kmuehlbauer:
This adds flags to code coverage. Introducing this the combined code coverage as well as the three different tests (doctests, unittests and notebooktests) are shown in the coverage reports.


kmuehlbauer:
So we can see that our code coverage for unittests is with 41% quite poor and need to be increased.

PR #170, opened 2017-08-21
ENH: IRIS reader

kmuehlbauer:
Adds Sigmet IRIS Reader using `numpy.memmap`.

This is work in progress. 

TODO:
- [x] add data type conversions (only sweep-data)
- [x] fully convert headers (eg. angles and datetimes)
- [x] strip NULL-characters from strings
- [x] add tests
- [x] add example file (see https://github.com/wradlib/wradlib-data/commit/c108492e81f12f73449f69f75b5d54573f7f4a25)
- [x] add docs
- [x] final PEP8-check


kmuehlbauer:
@heistermann I added one minor addition. Since all went well I'm going to squash-merge that soon.

Issue #171, opened 2017-08-25
Inconsistent shapes of arrays returned by different classes in the ipol module.


heistermann:
In case only one set of source values is passed to the `__call__` method, different classes in `wradlib.ipol` behave differently where they should behave consistently.

`Nearest` and `Idw` return 1-D (flat) arrays, `OrdinaryKriging` and `ExternalDriftKriging` return matrices of shape (numtargetpoints, 1). This causes problems e.g. when `OrdinaryKriging` is used as argument `Ipclass` in the `adjust` modules adjustment classes (`MemoryError` when matrix is multiplied with a flat array). This caused the issue that was raised here:

https://groups.google.com/forum/?fromgroups#!topic/wradlib-users/MXUVXONlhaY

We might have not explicitely caught this behavious so far because unit tests for `Idw` and `Nearest` are not implemented, and `OrdinaryKriging` has not been routinely used as an interpolation class when using classes from `adjust`.

Fixing this is ambivalent: The idea has always been that you can pass multiple source datasets in one 2-D array (num sourcepoints, num datasets) to the `__call__` methods. 

*Option 1:* Change `Idw` and `Nearest` so they return an array of shape (mum sourcepoints, 1) in case only one dataset is passed. This might, however, break several applications who have been using these classes so far. 

*Option 2:* Changing `OrdinaryKriging` to behave like e.g. `Idw` might break applications using `OrdinaryKriging`.

*Option 3*: The number of dimensions of the returned array should be the same as the number of dimensions of the input array. This might still break some applications: if people have so far been passing flat arrays to `Idw`, they will still get back flat arrays - so no problem. But if they have been passing a flat array to `OrdinaryKriging`, they will get back a flat array - which might still break their applications. 

My suggestion: Use Option 3, but not as a bug fix release, but with a new 1.0 release. What do you think, @kmuehlbauer ?

kmuehlbauer:
@heistermann If there is the slightest chance that things brake after the change, then we should take Option 3.

kmuehlbauer:
@heistermann Would also be a good opportunity to implement tests.

heistermann:
Yes, definetely. As pointed out, option 3 might also break applications, but we can explicetely state this with a new release. It is also more intuitive for those who have been using this with flat arrays. It would only require to ravel the out put from the Kriging classes in case only one dataset is used.

I will make the documentation more explicit in that regard, and, yes, implement tests for the other classes.

kmuehlbauer:
[...]

PR #172, opened 2017-08-25
Fixing dimensions issue in ipol and add tests,

heistermann:
This refers to issue #171. Fixes the inconsistent behaviour in ipol in case flat arrays are passed. Also adds some tests for ipol as well as adjust.


wradlib/ipol.py
```diff
@@ -200,6 +202,14 @@ def __call__(self, vals, maxdist=None):
         """
         Evaluate interpolator for values given at the source points.
 
+        You can interpolate multiple datasets of source values (`vals`) at
+        once: the `vals` array should have the shape (number of source points,
+        number of source datasets). If you want to interpolate only one set
+        of source values, `vals can have the shape (number of source points,
```
kmuehlbauer: Backtick missing at `vals`

wradlib/ipol.py
```diff
@@ -279,6 +289,14 @@ def __call__(self, vals):
...
```
kmuehlbauer: same as above

wradlib/ipol.py
```diff
@@ -365,6 +383,14 @@ def __call__(self, vals, fill_value=np.nan):
...
```
kmuehlbauer: same here

wradlib/ipol.py
```diff
@@ -642,6 +668,14 @@ def __call__(self, vals):
...
```
kmuehlbauer: and here

wradlib/ipol.py
```diff
@@ -797,6 +834,14 @@ def __call__(self, vals, src_drift=None, trg_drift=None):
...
```
kmuehlbauer: here also

kmuehlbauer:
@heistermann If we pull this in now, then it will be in the next release (0.10.2 or 0.11.0). Considering what you mentioned in #171, we wait until preparation of release 1.0.0, right?

heistermann:
👍

kmuehlbauer:
@heistermann I cherry-picked you changes and they are taken care of in PR #226. Closing.

PR #175, opened 2017-08-30
ENH: add arg to testrunner.py 

kmuehlbauer:
This enables `testrunner.py` to take a single string argument. Tests which `TestCase id()` (the name of the test) doesn't contain this argument string are removed from the TestSuites.

Say if you want to test only a specific test-function containing `rvp` you can do it like this:

```bash
(wradlib) user@machine:python testrunner.py -a rvp
rvp2dBZ (wradlib.trafo)
Doctest: wradlib.trafo.rvp2dBZ ... ok

----------------------------------------------------------------------
Ran 1 test in 0.002s

OK
test_rvp2dBZ (test_trafo.TransformationTest) ... ok

----------------------------------------------------------------------
Ran 1 test in 0.000s

OK
```
This makes local testing a lot easier.

### DIFF
--- a/testrunner.py
+++ b/testrunner.py
@@ -57,8 +57,10 @@ class NotebookTest(unittest.TestCase):
         self.nbfile = nbfile
         self.cov = cov
 
+    def id(self):
+        return self.nbfile
+
     def runTest(self):
-        print(self.nbfile)
         kernel = 'python%d' % sys.version_info[0]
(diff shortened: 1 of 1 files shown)


heistermann:
Bravo!

kmuehlbauer:
> Bravo!

Does that mean: Merge?

heistermann:
Yes!

PR #176, opened 2017-09-20
ENH: refining loaddata switch

kmuehlbauer:


kmuehlbauer:
@heistermann More to come, specifying `sweeps` and `moments` via `loaddata`.

Issue #178, opened 2017-09-26
Add argument for altitude offset in plot_rhi


heistermann:
`wradlib.vis.plot_rhi` should allow for an argument to specify the altitude of the radar antenna (e.g. `yoffset`) so that the actual beam altitude above m.s.l. can be represented. With `yoffset` defaulting to `0.` the API does not break.

kmuehlbauer:
resolved by #179

PR #179, opened 2017-09-26
Add altitude offset to plot_rhi

heistermann:
This PR addresses #178.


heistermann:
Just saw this does not work with `cg=True`, yet - curvilinear grid does not consider `yoffset`.

heistermann:
I don't quite see how the `cg` option could be tweaked to make this work. For the time being, we might force `yoffset=0.` in case `cg=True`. But that's not nice...

PR #185, opened 2017-10-16
Fix read radolan header

### LINKED ISSUE #184: Infinite loop when trying to read an empty RADOLAN file
If you read an empty (possibly also a corrupt) RADOLAN file, the execution gets stuck in an infinite loop in `wradlib.io.read_radolan_header(fid)`

This seems to be due to no check, if there is any input available. The code only checks for the EOT char to break the while(True) loop, but doesnt stop if it is an empty character.

kmuehlbauer:
fixes #184

### DIFF
--- a/wradlib/io/radolan.py
+++ b/wradlib/io/radolan.py
@@ -625,6 +625,9 @@ def read_radolan_header(fid):
     header = ''
     while True:
         mychar = fid.read(1)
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
@heistermann Just a minor issue... Added tests as well.

PR #187, opened 2017-10-17
Consider NaNs adequately in weighted composition

heistermann:
In case the quality / weighting grid for the composition are not masked with NaN values outside the radar domain, these areas should be masked based on the radar frids themselves. Otherwise, the resulting weights will be flawed.


kmuehlbauer:
@heistermann 👍

heistermann:
Merge rightaway? I think this does not have to wait until the next release.

kmuehlbauer:
Yes, pull it in...

PR #212, opened 2017-12-20
FIX: silence install.sh

kmuehlbauer:


kmuehlbauer:
If at first you don't succeed...

PR #233, opened 2018-03-07
ENH: add tests to georef

kmuehlbauer:
- [x] add tests to georef
- [x] remove duplicate functions from `misc`
- [x] several minor fixes

- [x] renamed `get_shape_coordinates` to `get_vector_coordinates`
- [x] renamed `get_shape_points` to `get_vector_points`

- [ ] usage of `georef.ogr_to_numpy` can be replaced by `georef.get_shape_points` (which will be renamed to `georef.get_vector_points`) **Update**: `get_vector_points` needs only 30 percent of the time compared to `georef_ogr_to_numpy`


kmuehlbauer:
@heistermann 
We can decide on the usage of `ogr_to_numpy` later. What I would do now is merge #231, #232 and this PR in exactly this sequence. I've planned to do this tomorrow early, so please make any remarks unti then.

PR #252, opened 2018-03-27
module rework atten removes a lot of functions and adds tests

heistermann:
renames functions, removes atten functions that are redundant or not used anymore, removes inconsistences.


kmuehlbauer:
@heistermann flake8 is mastering us all

PR #254, opened 2018-03-28
More module revisions

heistermann:
More to come... so far, this PR contains:

- moving functions from dp to util and add tests


kmuehlbauer:
@heistermann Do these effect notebooks?

PR #258, opened 2018-03-28
More module rev

heistermann:
Mainly speeding up the vpr tests and revising docstrings in vpr module.


heistermann:
Is not yet complete - more to come.

PR #259, opened 2018-03-28
FIX: rename ZonalStatsBase-derived classes

kmuehlbauer:


kmuehlbauer:
@heistermann I think `ZonalStatsPoint` and `ZonalStatsPoly` do match best into the overall class naming scheme. It's ZonalStats from Point source and ZonalStats from Polygon source.

PR #285, opened 2018-09-24
ENH: enable `ZonalDataBase` to take `DataSource`-object as parameter

### LINKED ISSUE #284: ZonalDataBase: Allow Datasource objects as 'src'/'trg'parameter
**Current Status**

 On initialization of a `ZonalDataPoint`/`ZonalDataPolygon` object we hand the given `src` and `trg` to the `DataSource`-class for initalization of a `DataSource`-object. If `src` is a string pointing to an `ESRI Shapefile`, the file will be used to initialize the `DataSource`. 

Although the loading of a pre-created shapefile is much faster than the generation of the `DataSource` it consumes quite some time. This is especially annoying if you are testing hundreds or thousands of targets against the same source in a loop.

**Proposed Status**

Create the `src` `DataSource` object before and allow `ZonalDataPoint`/`ZonalDataPolygon` (via `ZonalDataBase` class) to consume it as `src` parameter.

To make this work we need to think about the overall handling. The easiest way seems to provide `DataSource` with a `load_vector`-function and use it in the class __init__. The `_check_data` function can then be stripped from the `try/except` file handling and be used. Further the `ZonalDataBase` class needs checking which `src`/`trg`-parameter is given, numpy array, DataSource-object or filename. numpy goes to `_check_data ` and the other two to `load_vector`.

Positive side-effect: The `ZonalDataBase` `load_vector` function can be simplified.

Thoughts? I'll prepare a PR next week.

kmuehlbauer:
- add `load_vector`-method to `DataSource`
- enable `ZonalDataBase` to take `DataSource` objects as parameters
- minor change in `io.open_vector`
- adapted tests

- resolves #284

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -65,7 +65,7 @@ conda create -n $WRADLIB_ENV --yes pip python=$WRADLIB_PYTHON
 source activate $WRADLIB_ENV
 
 # Install wradlib dependencies
-conda install --yes gdal numpy scipy matplotlib netcdf4 h5py deprecation xmltodict coverage codecov nose nose-timer
+conda install --yes gdal numpy scipy matplotlib=2* netcdf4 h5py deprecation xmltodict coverage codecov nose nose-timer
 conda list
 
 # Install wradlib-data if not set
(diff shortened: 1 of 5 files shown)


kmuehlbauer:
Tests failing due to regression in matplotlib, see https://github.com/matplotlib/matplotlib/issues/12265.

Issue #294, opened 2018-10-11
Sigmet IRIS file reading loaddata kwdict option not working as expected


jorahu:
wradlib.io.iris.read_iris documentation example:

loaddata (bool | kwdict) –
If true, retrieves whole data section from file. If false, retrievs only ingest_data_headers, but no data. If kwdict, retrieves according to given kwdict:

loaddata = {'moment': ['DB_DBZ', 'DB_VEL'], 'sweep': [0, 3, 9]}

The **'sweep'** parameter name is incorrect in source code, it is referred to as **'data'** in the try clause below:

```
def get_data(self):
        """Retrieve all sweeps from file.
        """
        dt_names = [d['name'] for d in self.data_types]
        rsweeps = range(1, self.nsweeps + 1)

        loaddata = self.loaddata
        try:
            sweep = loaddata.copy().pop('data', rsweeps)
            moment = loaddata.copy().pop('moment', dt_names)
        except AttributeError:
            sweep = rsweeps
            moment = dt_names
```

The consequence was that even after specifing the 'sweep' parameter at the loaddata kwdict it read all sweeps every time anyways (usually taking around 15 seconds for 1 file) . If I tried 
loaddata = {'moment': ['DB_DBZ'], **'data'**: [1]}
then everything worked as expected, it read only first sweep and DB_DBZ field (now taking around 2 seconds per file). (Also the example was for sweep 0, but actual sweep counting starts with 1 as seen from rsweep above.)

kmuehlbauer:
@jorahu Thanks for reporting this. I'll have a look into it the next day. Which source file did you use?

jorahu:
IRIS RAW file.

kmuehlbauer:
@jorahu Is it possible to make the file available somewhere? If it is not too big you can send it per email. You can find my credentials via my github bio.

kmuehlbauer:
@jorahu Thanks for reaching out and sending the data file. I can reproduce the issue. This is a bug, which only surfaces if the `loaddata` is used as `kwdict` and then it might only get spotted if the loading time is watched.

The shown `get_data` method is the correct place for the fix. 

The line should read: `sweep = loaddata.copy().pop('sweep', rsweeps)`. I'll also polish the docstring a bit, since there is no sweep `0` in Iris data. 

@jorahu Thanks again for filing that issue, this is much appreciated. If you have any other questions or remarks concerning the sigmet/iris reader or wradlib let us know. Greetings to Estonia!

[...]

PR #295, opened 2018-10-12
FIX: align docstring with actual code and use `sweep`

kmuehlbauer:
- align docstring with actual code and use `sweep`
- do not use `0` for sweep number in docstring
- fix tests


kmuehlbauer:
closes #294

PR #296, opened 2018-10-16
ENH: add function to retrieve radiosonde data 

kmuehlbauer:
- add function to retrieve radiosonde data to `io.misc`, 
- add test


kmuehlbauer:
I thought about the name too, but did not find any better. `fetch_radiosonde` was another name.

kmuehlbauer:
@heistermann `get_radiosonde` seems good enough. Will merge if the CI is happy.

PR #299, opened 2018-10-30
MNT: unpin matplotlib

kmuehlbauer:


kmuehlbauer:
Matplotlib 3.0.1 fixes the pcolormesh problem, but adds a regression which makes curve linear plots break in notebooks. So we need to stick to latest matplotlib 2 version.

heistermann:
Thanks a lot for fixing this!! ANd yes, I agree it is time to drop 2.7.

PR #302, opened 2018-11-08
Odim

egouden:
Here it is !
Please note that I use pytest for testing.
Unfortunately I can't run the test since I am having trouble with the version module.
I am open to suggestions on how to proceed from now.

# Update
Data sets for testing: 
http://eumetnet.eu/wp-content/uploads/2017/04/ENMI_ODIM22_H5.zip
http://eumetnet.eu/wp-content/uploads/2017/04/EHDB_ODIM22_H5.zip

ODIM standart V2.2
http://eumetnet.eu/wp-content/uploads/2017/01/OPERA_hdf_description_2014.pdf

ToDo:
- [ ] `count_odim_datas` returns only count in `dataset1` assuming all datasets have the same count of datas inside. But every dataset can contain an arbitrary number of parameters (datas).
- [ ] offset and gain is only read for `dataset1`. This might change for different datasets, so we need to fix this
- [ ] this affects also the value decoding, since values from datasetN are decoded with offset and gain from dataset1
- [ ] h5py writes strings as H5T_STR_NULLPAD whereas the ODIM spec explicitely declares the strings to be H5T_STR_NULLTERM. 
- [ ] DATASET "data" ATTRIBUTE `CLASS` and `VERSION`/`IMAGE_VERSION` missing in output file (only for 8-bit unsigned depth (`uchar`) 
- [ ] keep information of the DATASET DATATYPE (like `H5T_STD_U8LE`) before applying gain and offset, to reverse this correctly on write. If created from scratch, we need to add  a datatype
- [ ] using `astype` for downcasting (float->int) creates problems: 51.999999999 is casted to 51
- [ ] Is the `__eq__` override really needed for the ODIM part? If it's just there for testing we can apply testing framework from numpy and dictdiffer
- [ ] tbc


kmuehlbauer:
@egouden First of all, thank you for your efforts in implementing ODIM module.

Travis is breaking in the first place because of the PEP8 Errors. You would need fix these, or I can make a PR to your branch if. There are also imports (`rmiradlib`) which need to be sorted aut. I'll have a closer look.

kmuehlbauer:
@egouden Just commited minor changes to your branch, to get things running within the CI.

kmuehlbauer:
@egouden @heistermann If you open a can of worms, the only way to re-can them is to use a larger can.  *We'll have to use a larger can now.*

@egouden This is fantastic work. There are some things we have to consider and fix. I'll have a list in the PR opener. I also linked two example datasets, which we should use for testing.

wradlib/io/odim.py
```diff
@@ -0,0 +1,674 @@
...
+            vald = []
+            for data in datas:
+
+                val = self.filehandler[dataset][data]['data']
+                if slicing is None:
+                    val = val[:]
+                else:
+                    val = val[slicing]
+
+                vald.append(val)
```
kmuehlbauer: The problem with this numpy array within a list is, that the different 'data' sets might have different datatypes, so squeezing them into one array will fail then.

kmuehlbauer: I have to take back that, since the the different datasets will be upcasted to the minimum type required to hold the objects. Anyway, we have to think if upcasting is really wanted here, or if it might be better to leave the arrays with their respective datatypes.

kmuehlbauer:
The string issue can be solved like this (see [stackoverflow](https://stackoverflow.com/questions/38267076/how-to-write-a-dataset-of-null-terminated-fixed-length-strings-with-h5py)):

```python
tid = h5py.h5t.C_S1.copy()
tid.set_size(len(value) + 1)
H5T_C_S1_NEW = h5py.Datatype(tid)
group.attrs.create(key, value, dtype=H5T_C_S1_NEW)
```

kmuehlbauer:
@egouden Many thanks for all the work you did with this implementation. After wrapping my head around everything and talking to colleagues all over the world (thanks especially @DanielMichelson) I came up with the implementation in #311. Please have a look there, if you are interested. I'm going to close this PR.

egouden:
Hi Kai,

This looks interesting apart from being completely different from my
approach ;-)
I understand it as a high level but also general implementation providing
some extra features.
First thing I will check (compare) is the speed to read the values.
But I got my first issue:

filename = 20190220094000.rad.bewid.pvol.dbzh.scanz.hdf
xr.OdimH5(filename)["sweep_1"]["DBZH"].values
RuntimeError: NetCDF: Not a valid ID

Best regards,

Edouard


Le ven. 15 févr. 2019 à 11:45, Kai Mühlbauer <notifications@github.com> a
écrit :

kmuehlbauer:
@egouden Can you send me the file? I made some changes to sources, do not know if I pushed them already.

egouden:
Is the file not attached to my previous mail?

Le mer. 20 févr. 2019 à 14:39, Kai Mühlbauer <notifications@github.com> a
écrit :

kmuehlbauer:
@egouden No, I think github is removing the attachments.

egouden:
Just solved it.
This works:
>>> pvol = xr.OdimH5("20190220094000.rad.bewid.pvol.dbzh.scanz.hdf")
>>> pvol["sweep_1"]["DBZH"].values



Le mer. 20 févr. 2019 à 14:45, Kai Mühlbauer <notifications@github.com> a
écrit :

kmuehlbauer:
@egouden Got the same error. You need to have the object like this:

filename = 20190220094000.rad.bewid.pvol.dbzh.scanz.hdf 
ds = xr.OdimH5(filename)
ds["sweep_1"]["DBZH"].values

Issue #301, opened 2018-11-02
Reading data from IRIS ingest files


aschueth:
I would like to be able to read data from binary IRIS ingest files, that are not in the same format as the product_raw files, etc... The format for them is a main header file for that sweep, and then a different file for every radar moment. A sample dataset can be found here: https://drive.google.com/drive/folders/18w3QBu246BoNvPvCTGN4cCaz1I5XwJZR?usp=sharing

While the data linked are PPIs, I also have RHIs that I would like to read in as well.

kmuehlbauer:
@aschueth Thanks for opening this issue and providing the ingest data files.

For implementation we will use numpy memmap to get hold of the files contents (as it is done already for the normal sigmet/IRIS files).

Unfortunately we can't use the record approach since the ingest files doesn't have record structure. But we can use all the implemented decoding mechanism.

From my perspective it would be sufficient to create reader functions for the ingest header file and for the ingest data files. So we do not need to put that much magic inside.

I'll link the implementation PR here early next week.

@aschueth Can I contact you offlist regarding a working notebook prototype?

aschueth:
Absolutely

aschueth:
Absolutely! My email can be found at my account homepage

kmuehlbauer:
@aschueth I just sent you an email with a working prototype, so you can start reading the ingest data and metadata.

Since the ingest data has no record structure reading is a bit different from the other sigmet files. It might be necessary to re-think the class structure a bit. Just notifying you, @heistermann, that I might contact you regarding this.

kmuehlbauer:
@aschueth If you find the time you might test against my [feature branch](https://github.com/kmuehlbauer/wradlib/tree/sigmet-redesign). See PR #303.

aschueth:
I tested the feature branch, I might be missing something but how do I read
in the data? In the first notebook you sent me there was a function sweep
that returned the field data. Additionally, how can I access the nyquist?

kmuehlbauer:
@aschueth Just sent you an updated notebook. I had to make minor changes to be compatible with the structure of the other classes.

kmuehlbauer:
@aschueth #303 is pulled in. Expect the changes to be available from wradlib version 1.4.0.

mustafaalicelik:
Hi,

I have the same issue about viewing IRIS ingest files. Can you send me the prototype too?

kmuehlbauer:
[...]

Issue #180, opened 2017-09-29
Sigmet Reader Todo List


kmuehlbauer:
The Sigmet IRIS Reader is in general capable of reading all types of product files:

* Ingest Data Files
* RAW product (polar data)
* cartesian products (MAX, TOPs etc)
* other products (FCAST, SLINE, WARN, WIND)

The overall class-structure is fixed, but this is what is yet to be done:

* [ ] thorough documentation
* [ ] decoding of data review (is all data converted correctly using the given sigmet_data_type)
* [ ] add necessary decoding functions
* [ ] mask nodata and area not scanned (or add array-masks)
* [ ] add necessary data structs (there are still some left)
* [ ] check class-properties (getters/setters)
* [x] possibly add class-structures for headers  PR #303 
* [ ] check for possible simplifications in functions
* [x] fix `DB_KDP` decoding (division by wavelenght) #195 
* [ ] research windows issue, where unusual high kdp values are extracted at some bins
* [x] decode iris DB_XHDR as numpy structured array #488 
* [x] add rainrate decoder to iris reader #428
* [x] make RAW files lazy loading and improve reading #520 
* [x] add xarray backend for RAW files #520 

See the [source code](https://github.com/wradlib/wradlib/blob/master/wradlib/io/iris.py).

kmuehlbauer:
This still lacks sigmet test files of all products. If there is someone who can provide these data files to be made openly available, just contact wradlib maintainers.

kmuehlbauer:
The development continues in xradar-package https://github.com/openradar/xradar for polar source data.

Cartesian data is handled here until decision if xradar will take care of cartesian data too.

PR #303, opened 2018-11-09
ENH: Sigmet redesign

kmuehlbauer:
- Implementation of #301 
- Follow up on #180

This is WIP!


kmuehlbauer:
@ckaradavut Thanks for pointing out the issue with `DB_FLIQUID2`. I commited the patch from your branch. After checking I found some minor glitch (operator precendence), which is fixed with this PR.

Also you might find it useful to read in IRIS Ingest Files, which is implemented in this PR (besides other things)

PR #312, opened 2019-02-15
FIX: coords, AttributeError

kmuehlbauer:
assign coords (azimuth, elevation) for cfradial, fix exception for AttributeError


kmuehlbauer:
This happens if you move from a development system to a production system without proper testing. 

I expect more fixes to be included in the codebase. Will create a special issue/project for tracking this.

PR #313, opened 2019-02-18
MNT: correct use of semver

kmuehlbauer:
The development versions need to have the upcoming versions MINOR-Level.

- current stable: 1.2
- current patch: 1.2.1
- upcoming: 1.3.0

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ netCDF4
 gdal
 deprecation
 xmltodict
+semver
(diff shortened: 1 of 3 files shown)


kmuehlbauer:
The workflow depends on using the git-repository including the tag/hash-information. Using `git describe --tags` the correct version is retrieved and written to `version.py` at install time. This complete package (including version.py) is bundled for the download from PyPI.

If no git information is available or version.py is not present the version is considered 'MAJOR.MINOR.PATCH-unknown'  as retrieved from `setup.py`. To keep it simple the next upcoming `MINOR` value is used within setup.py.

PR #315, opened 2019-02-22
Odim/CfRadial clean-up

kmuehlbauer:
This PR cleans-up classes (move static stuff out) and adds xarray to wradlib dependencies.


kmuehlbauer:
The radiosonde retrieval is very flaky. If there is a HTTPError it is catched and the test is skipped.

PR #316, opened 2019-03-07
ENH: Xarray plotting

kmuehlbauer:
Wraps PPI and RHI Plotting Routines around xarray DataArray .

- works transparent on current notebooks/ test cases /code base (might be corner cases)
- adds xarray DataArray creation from spherical coordinates/proj
- adds wradlib xarray DataArray Accessor for plotting
- adds necessary attributes to ODIM/CfRadial reader functions to be able to use plotting
- introduces cartopy mapping possibilities (optional)
- simplifies curvelinear grid creation (PPI/RHI discrimination via `rot` and `scale` keywords)
- deprecates `autoext`, `refrac` and `cg` keywords (managed by xarray and re/ke)
- added/fixed tests


kmuehlbauer:
@heistermann May I bother you to have a look into this? There might be some glitches left, the test suites are working as expected. I also introduced a notebook-testsuite running the wradlib-notebooks master branch as extension of the two normal test suits in #318. So in wradlib-master as well as in this PR everything is green.

I already have notebooks which I'll add/change when this is pulled in. After that I'll give the notebooks a refresh, polish the docs and finally release wradlib 1.3 in the next days but latest next week. So please add anything which need to be pulled in with this version.

PR #329, opened 2019-04-26
MNT: Updates import for MutableMapping from collections to remove warnning. abc was added.

zssherman:
Adds .abc in collection to remove warning when importing wradlib.


kmuehlbauer:
@zssherman Let us know, if this change should go into a patch release.

zssherman:
@kmuehlbauer From what I'm seeing it just raising the single warning for now, but when python 3.8 is released, it will throw an error. I think its fine for now.

kmuehlbauer:
Yeah, we need to make sure to have a release out, if python 3.8 hits conda-forge.

PR #331, opened 2019-04-29
FIX: correct texture processing

kmuehlbauer:
- correct nan-handling in texture
- improve performance using numpy views


kmuehlbauer:
@overeem11 Would be good to make the [test](https://github.com/wradlib/wradlib/blob/239efd091e1492a6a5c004271a109a22853e4072/wradlib/tests/test_dp.py#L51) a real test to test against these kind of problems. WDYT?

PR #332, opened 2019-05-09
Read radolan from piped stdin

PPazderski:
I'd like to read a radolan file from stdin but am prevented from the

>  rewind, just in case...

Unfortunately stdin connected to a pipe is not seekable.

Apart from not seekable, if the user pass in a file handle it is his responsibility to position it correctly. The radolan data may not be contained at the start of the file handle.


kmuehlbauer:
Ping @heistermann

Issue #326, opened 2019-04-11
xarray reader performance


egouden:
Climate studies imply the processing of decades of radar data.
It is therefore important to read the data quickly.
After diging into the code and looking at the profiling results (see below), I suggest to add the following features to the xarray interface:
- option to get a subset of the data (imagine you want to focus on a city)
- option to not decode the data
- option to not get coordinate information
- option to not get the xyz coordinates
- option to not read any extra metadata
- option to read only recommended metadata

I guess this could be easily implemented by dividing the big assign functions into smaller ones.
Attached is a pull request but just to highlight the issues(there is sitll a call to decode_cf I cannot find).
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      2/1    0.000    0.000    0.808    0.808 {built-in method builtins.exec}
        1    0.000    0.000    0.808    0.808 <string>:1(<module>)
        1    0.000    0.000    0.806    0.806 test_io.py:1048(test_read_odim)
        3    0.018    0.006    0.802    0.267 xarray.py:1191(__init__)
        2    0.024    0.012    0.781    0.390 xarray.py:1227(assign_data_odim)
      126    0.002    0.000    0.260    0.002 conventions.py:412(decode_cf)
      126    0.003    0.000    0.156    0.001 conventions.py:351(decode_cf_variables)
      207    0.004    0.000    0.152    0.001 merge.py:392(merge_core)
      380    0.007    0.000    0.150    0.000 conventions.py:245(decode_cf_variable)
      171    0.001    0.000    0.132    0.001 common.py:295(assign_coords)
     9343    0.018    0.000    0.127    0.000 variable.py:239(__init__)
      457    0.007    0.000    0.126    0.000 common.py:171(__getattr__)
       19    0.011    0.001    0.118    0.006 polar.py:30(spherical_to_xyz)
      627    0.008    0.000    0.109    0.000 merge.py:172(expand_variable_dicts)
     2860    0.013    0.000    0.105    0.000 variable.py:41(as_variable)
      171    0.001    0.000    0.100    0.001 coordinates.py:89(update)
       59    0.000    0.000    0.091    0.002 dataset.py:3001(assign)
       78    0.000    0.000    0.089    0.001 dataset.py:2507(update)
       78    0.002    0.000    0.088    0.001 merge.py:559(dataset_update_method)
      743    0.002    0.000    0.085    0.000 dataset.py:960(__getitem__)
```

kmuehlbauer:
[...]

PR #333, opened 2019-05-14
ENH: parametrize xarray loading

kmuehlbauer:
- add keywords for parametrized reading of OdimH5 data, 
- add more OdimH5 moments

closes: #326


kmuehlbauer:
@egouden This adds the keywords as laid out in #326. Need to add the GAMIC stuff. Would be nice, if you can give it a try already.

PR #334, opened 2019-05-17
ENH: use dr and cpa in classify_echo_fuzzy 

kmuehlbauer:
- add calculation of `depolarization` ratio to `wradlib.dp`
- enhance/rewrite `wradlib.clutter.classify_echo_fuzzy`


kmuehlbauer:
@overeem11 Thanks for this contribution. I took the chance to rewrite the fuzzy function to be more readable. Please check if this works correct for your use-case.

PR #342, opened 2019-06-04
Create .gitignore file (Issue #339)

### LINKED ISSUE #339: Add .gitignore
To prevent comitting unwanted files to wradlib codebase a `.gitignore` file should be added.

nschaffner:
I have created a `.gitignore` file for this repository based on the GitHub Python.gitignore template. Please let me know if there are any changes you would like for me to make to the file to better serve this project.

closes #339

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,23 @@
+# ignore everything
+*
+
+# whitelist block
+# whitelist root .gitignore
+!/.gitignore
+
+# whitelist .py, .f-files and .sh files
+!*.py
+!*.f
+!*.sh
+
+# whitelist files under `scripts`
+!/scripts/
+!/scripts/*
+
+# whitelist files under `wradlib`
+!/wradlib/
+!/wradlib/*
+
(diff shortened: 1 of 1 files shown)


kmuehlbauer:
@nschaffner 

Thanks for looking into this. Could we try to add something like this to the top?

```
# ignore everything
*

# whitelist block
# whitelist root .gitignore
!/.gitignore

# whitelist .py, .f-files and .sh files
!*.py
!*.f
!*.sh

# whitelist files under `scripts`
!/scripts/
!/scripts/*

# whitelist files under `wradlib`
!/wradlib/
!/wradlib/*
```

This will make sure, that only the current folders `scripts` and `wradlib` as well their subfolders are taken into account and only the python, fortran and shell files below. If we want to add other files later we would need to adapt `.gitignore`.

We can take blacklist part to make sure that in all subfolders only wanted files are taken into account.

nschaffner:
I will add the section you have included to the top of the .gitignore file and push the changes.

PR #345, opened 2019-06-04
Feature/srtm

egouden:
Basic access to srtm data.

### DIFF
--- a/wradlib/io/dem.py
+++ b/wradlib/io/dem.py
@@ -18,13 +18,13 @@ get_srtm
 
 """
 
-import numpy as np
 import os
 
 from osgeo import gdal
 import requests
 
 import wradlib.util
+
 
 def download_srtm(filename, destination=None, version=2, resolution=3,
                   region="Eurasia", token=None):
@@ -43,7 +43,7 @@ def download_srtm(filename, destination=None, version=2, resolution=3,
     resolution : int
         resolution of SRTM data (1, 3 or 30)
     region : str
-        name of the region for SRTM version 2 only: 
+        name of the region for SRTM version 2 only:
         Africa, Australia, Eurasia, Islands, North America, South America
     token : str
         filename with authorization token (required for version 3)
@@ -52,22 +52,21 @@ def download_srtm(filename, destination=None, version=2, resolution=3,
 
     if version == 2:
         website = "https://dds.cr.usgs.gov/srtm/version2_1"
-        source = "%s/SRTM%s/%s" %(website, resolution, region)
+        source = "%s/SRTM%s/%s " % (website, resolution, region)
     if version == 3:
-        website = "http://e4ftl01.cr.usgs.gov/MEASURES" 
-        source = "%s/SRTMGL%d.003/2000.02.11" %(website, resolution)
-    url = "%s/%s" %(source, filename)
+        website = "http://e4ftl01.cr.usgs.gov/MEASURES"
+        source = "%s/SRTMGL%d.003/2000.02.11" % (website, resolution)
+    url = "%s/%s" % (source, filename)
 
     headers = None
     if token is not None:
-        headers = {'Authorization': 'Bearer %s' %(token)}
+        headers = {'Authorization': 'Bearer %s' % (token)}
     try:
         r = requests.get(url, headers=headers, stream=True)
         r.raise_for_status()
     except requests.exceptions.HTTPError as err:
         print(err)
         return
-    
     if destination is None:
         destination = filename
     with open(destination, 'wb') as fd:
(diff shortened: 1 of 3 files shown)


wradlib/io/dem.py
```diff
@@ -0,0 +1,142 @@
...
+
+"""
+
+import numpy as np
+import os
+
+from osgeo import gdal
+import requests
+
+import wradlib.util
```
kmuehlbauer: please use relative import here and specify function: `from ..util import get_wradlib_data_path `

wradlib/io/dem.py
```diff
@@ -0,0 +1,142 @@
...
+
+    headers = None
+    if token is not None:
+        headers = {'Authorization': 'Bearer %s' % (token)}
+    try:
+        r = requests.get(url, headers=headers, stream=True)
+        r.raise_for_status()
+    except requests.exceptions.HTTPError as err:
+        print(err)
+        return
```
kmuehlbauer: We need to think about a better solution here. Just swallowing the exception here doesn't seem the right way IMHO.

Simplest thing would be to add a specifying wradlib message to the error message and re-raise.

wradlib/io/dem.py
```diff
@@ -0,0 +1,142 @@
...
+    extent : str
+        lonmin, lonmax, latmin, latmax
+    version : int
+        srtm version (2 or 3)
+    resolution : int
+        resolution of SRTM data (1, 3 or 30)
+    merge : bool
+        True to merge the tiles in one dataset
+    download : dict
+        download options (see downloard_srtm)
```
kmuehlbauer: Typo

wradlib/io/dem.py
```diff
@@ -0,0 +1,142 @@
...
+            filelist.append(georef)
+        for longitude in range(max(lonmin, 0), lonmax+1):
+            georef = "N%02gE%03g" % (latitude, longitude)
+            filelist.append(georef)
+    if version == 3:
+        filelist = ["%s.SRTMGL%s" % (f, resolution) for f in filelist]
+
+    filelist = ["%s.hgt.zip" % (f) for f in filelist]
+
+    wrl_data_path = wradlib.util.get_wradlib_data_path()
```
kmuehlbauer: strip `wradlib.util`, if we import function directly

wradlib/io/dem.py
```diff
@@ -0,0 +1,142 @@
...
+    if not os.path.exists(srtm_path) and download is not None:
+        os.makedirs(srtm_path)
+    dems = []
+    for filename in filelist:
+        path = "%s/%s" % (srtm_path, filename)
+        if os.path.exists(path):
+            dems.append(path)
+            continue
+        if download is not None:
+            print("Downloading %s" % (filename))
```
kmuehlbauer: I'm thinking about starting to use some `logger` module for this kind of messages. We make use of this only in `atten.py` but it would make sense to distribute this through the library.

wradlib/io/dem.py
```diff
@@ -0,0 +1,142 @@
...
+        for longitude in range(max(lonmin, 0), lonmax+1):
+            georef = "N%02gE%03g" % (latitude, longitude)
+            filelist.append(georef)
+    if version == 3:
+        filelist = ["%s.SRTMGL%s" % (f, resolution) for f in filelist]
+
+    filelist = ["%s.hgt.zip" % (f) for f in filelist]
+
+    wrl_data_path = wradlib.util.get_wradlib_data_path()
+    srtm_path = "%s/geo" % (wrl_data_path)
```
kmuehlbauer: Sorry if this might be nitpicking, I would use `os.path.join` for all path-related combining.

wradlib/io/dem.py
```diff
@@ -0,0 +1,142 @@
...
+    if not os.path.exists(srtm_path) and download is not None:
+        os.makedirs(srtm_path)
+    dems = []
+    for filename in filelist:
+        path = "%s/%s" % (srtm_path, filename)
```
kmuehlbauer: use `os.path.join`

wradlib/io/dem.py
```diff
@@ -0,0 +1,142 @@
...
+            continue
+        if download is not None:
+            print("Downloading %s" % (filename))
+            download_srtm(filename, path, version, resolution, **download)
+            if os.path.exists(path):
+                dems.append(path)
+
+    dems = [gdal.Open(d) for d in dems]
+
+    if not merge:
```
kmuehlbauer: We might switch the last lines with this:
```python
if merge:
    dems = gdal.Warp('', dems, format='MEM'

return dems
```

wradlib/tests/test_io.py
```diff
@@ -1150,5 +1151,41 @@ def test_cfradial_odim_roundtrip(self):
                                    cf3.root.time_coverage_start)
 
 
+class DemTest(unittest.TestCase):
+
+    def test_get_srtm(self):
+        targets = ["N51W000", "N52E000", "N51E001",
```
kmuehlbauer: `targets` is not used within the test. Do you want to extend this?

kmuehlbauer:
@egouden Good work! This is a very neat feature, which comes in handy in several places (beamblockage, plotting dem-maps etc.) 

As I understood, this reads the files from the `/geo` subdirectory in the wradlib-data repo (or where the WRADLIB_DATA-variable is pointing to), If the files are not there it will download them, right?

I added comments with some suggestions. Also please add the the main function `get_srtm`to `wradlib.io.__init__.py` so that it can be called like this `wradlib.io.get_srtm`.

wradlib/io/dem.py
egouden: OK for relative import.
But what is the rationale for specifying functions?
I got stuck several times with circular dependencies georef->io->hdf->zonalstats->georef
Not specifying functions solve this.
Is it not a general recommendation not to use this?
I think it is also easier to maintain

wradlib/io/dem.py
egouden: Actually it has to pass if the HTTP error is 404 because it is expected that "sea" files are missing. Other HTTP errors should reraise.

wradlib/io/dem.py
egouden: Yes a logger would be better (although the benefits of logging in general should be discussed).
Could we simply remove the print now?

wradlib/io/dem.py
egouden: OK

wradlib/io/dem.py
kmuehlbauer: I agree. The `from package import module` is only suggested for submodules not functions, see this module-doc [here](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package). 

And you are of course right, that is makes the code more self describing and maintainable.

wradlib/io/dem.py
egouden: This was made for readability...

wradlib/tests/test_io.py
egouden: Oups...

egouden:
What about the coverage of the download function which seems difficult to test?

wradlib/io/dem.py
kmuehlbauer: > What about the coverage of the download function which seems difficult to test?

We could try to test, but decorate it with "knownfail"?

kmuehlbauer:
> What about the coverage of the download function which seems difficult to test?

We could mark it as knownfail.

wradlib/io/dem.py
kmuehlbauer: Yes, then remove it.

egouden:
Manually adding submodules functions to the module __init__ is painful to me ;-).
Is it really useful?
Is there no hack to take all functions from submodules?

kmuehlbauer:
> Manually adding submodules functions to the module **init** is painful to me ;-).
> Is it really useful?
> Is there no hack to take all functions from submodules?

We will never really know in which direction the package evolves. So I really like to see that only functions get added to the parent-module namespace which we consider to be used by users.

Think just about using dot-completion using for instance with `wradlib.io.`. If we would allow _all_ functions, objects etc to live there it would be useless. So I think we just add this one line to `io/__init__py`. 

```
from .dem import get_srtm
```

egouden:
What about commit history and squashing?

kmuehlbauer:
We can either use github for squash merge, or you squash your commits into one and we do a rebase merge afterwards.

From my pov it is not necessary to add all commits by merge commit.

egouden:
It feels good to contribute ;-)

egouden:
I also tested the secure download of version 3 at 1 degree resolution with success.

egouden:
When I run the test script on my laptop, it crashes due to the many figures...

kmuehlbauer:
I did not had such issues in the past. Is there something we can do about?

egouden:
It works now but it is quite slow.
Do you perform any manual check on the figures?
We could perhaps force the use of the agg renderer for the tests?

kmuehlbauer:
No, we don't do checks, just looking that nothing breaks... Agg-rendering seems a good approach.

kmuehlbauer:
But this can be done in another PR :grinning :

wradlib/io/dem.py
```diff
@@ -0,0 +1,147 @@
...
+        return(dems)
+
+    dem = gdal.Warp('', dems, format='MEM')
+
+    return(dem)
```
kmuehlbauer: @egouden If we decide to leave the double return, please remove the brackets here.

kmuehlbauer:
@egouden Just one minor thing, then we can squash merge this in. Let me know, when you're happy merging.

kmuehlbauer:
@egouden LGTM! I'll merge this when the CI is happy, unless you stop me for some reason.

kmuehlbauer:
@egouden Thank you for the contribution! We'll have to update some notebooks then...

PR #357, opened 2019-07-05
Hotfix/gdal3

### LINKED ISSUE #340: GDAL >= 3 using PROJ.4 >= 6
Large changes have been made to the GIS-stack in names of GDAL and Proj.4.

XRef:
- https://gdal.org/development/rfc/rfc73_proj6_wkt2_srsbarn.html
- https://gdalbarn.com/

A hotfix might be needed soon, to not break things when users update their gdal/proj.4

egouden:
I am using archlinux which just got the update to gdal3.
Here is the migration guide:
https://github.com/OSGeo/gdal/blob/release/3.0/gdal/MIGRATION_GUIDE.TXT
So I tried to fix the issues detected by the tests.
I think the most important changes have been made.
However some of the remaining issues are related to:
- failed wkt equality tests: I think an exact match of WKT is not relevant
- German projections not properly handled in coordinate transformation

I need your help to finish this hotfix ;-)

fixes #340


kmuehlbauer:
@egouden Thanks for taking the lead here! Let me try to get this working locally.

kmuehlbauer:
@egouden This can get really complicated. I'll setup a gdal3 environment within the CI.

kmuehlbauer:
@egouden Could you have another look? I think all issues are resolved now. There are some things with differences between gdal2 vs gdal3 behaviour (but this might also be cause by libproj). We can track this down when it becomes urgent, imho.

kmuehlbauer:
@egouden @heistermann If there are no objections, I'll pull this in tomorrow morning.

wradlib/georef/projection.py
```diff
@@ -83,33 +115,39 @@ def create_osr(projname, **kwargs):
                    'AXIS["Latitude", NORTH]],'
                    'PROJECTION["polar_stereographic"],'
                    'PARAMETER["central_meridian", 10.0],'
-                   'PARAMETER["latitude_of_origin", 60.0],'
+                   'PARAMETER["latitude_of_origin", 90.0],'
```
cchwala: @kmuehlbauer @egouden Is this really correct? It though the old value of 60.0 is in line with the RADOLAN documentation.

kmuehlbauer: @cchwala I think this is correct. `latitude_of_origin` is the 90 deg (pole) whereas the 60 deg references the `standard_parallel` and the 10 deg the `central meridian`. `standard_parallel` seems not to be a parameter of the polar stereographic projection, afaik. 

The erroneous numbering didn't have an effect before maybe due to silent correction. With new gdal 3 this surfaced.

cchwala: Thanks for the quick reply. I just asked because I am tracking down a problem that the RADOLAN latitudes grid produced by a recent wradlib is slightly different from the grid produced with a former version. But after looking in more detail on my side, I found that my "recent wradlib" was still older than this PR. Unfortunately I do not know which "older" version I used. My guess is that it is a pyproj or gdal problem anyway. Nevertheless, I thought this 60 deg vs 90 deg difference could be a problem. Thanks anyway.

kmuehlbauer: @cchwala please do not hesitate to open an issue, if you think there is something fishy going on.
Please also note, that there is the trigonometric fallback available for the grid functions. They do not use gdal/proj.

cchwala: Thanks for the hint with the trigonometric fallback. I just checked and the results show a maximal difference to the gdal ones of 5e-14 degree. That should be sufficiently accurate for weather radar data ;-).

I will track down the problem on my side and open a PR if required.

kmuehlbauer: @cchwala The latitude thing didn't let me go, so I've looked it up. We use the `k0` values (aka `scale_factor`, `k0 = (1 + sin(lat_ts)) / 2`) instead of `standard_parallel`/`lat_ts` in our calculations. The radolan projection assumes a spherical earth. We transform this projection into WGS84 (which is using and ellipsiod model). This might explain the differences to the trigonometric fallback (which is pure spherical calculations).

cchwala: Thanks for looking into the details. My initial "problem" that was only visible in the latitude values, was not due to using trigonometric, though. I always used the default way via gdal.

I still have to track down the cause for the small differences when using older and newer wradlib or gdal/proj.

kmuehlbauer: Hmm, there where many fixes/changes in PROJ as well as in GDAL. We tried to follow these changes as best as possible. I remember that we had to change values in our tests to accommodate the altered behaviour.

cchwala: I had a look and, as you wrote, there have been many changes in the relevant projection code. It is not worth the effort to track this problem down since it is very unique to my usecase where I concatenate yearly RADOLAN NetCDFs over time, which only works seamlessly when the x-y and lat-lon coordinates are the same. My simple fix is, just using a predefined RADOLAN grid and not a dynamically derived one from wradlib. I could have used your suggestion with the trigonometric fallback, but to be 100% sure that nothing affects my RADOLAN NetCDFs in the future I think it is best to use the fixed predefined grid.

Thanks for your support!

kmuehlbauer: Thanks for the update. Predefined grids looks like the best approach to conquer this.

PR #374, opened 2019-08-05
add missing dependency 'requests'

s-m-e:
v1.5 is missing a dependency in its `requirements.txt`: `requests`

It's being used [here](https://github.com/wradlib/wradlib/blob/847a3c583081747394d0e55494506b815c5af2e3/wradlib/io/dem.py#L26).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ gdal
 deprecation
 xmltodict
 semver
+requests


egouden:
It was just easier to use Requests which is a kind of standard.
But one could use urllib/urllib2 to limit the dependency list if we don't plan many other http requests in the future.

kmuehlbauer:
@egouden I see the point. So if using `urllib` is too tedious we should keep using `requests`. The only problem would be adding the custom header, right?

kmuehlbauer:
@egouden Thinking a bit more about it, I'm already convinced. `requests` is already in the minimal conda python, so we should not bother about it. I'll pull the changes in after merging #381 and rebase.

s-m-e:
`requests` is the [7th most used pypi package](https://hugovk.github.io/top-pypi-packages/). There is nothing wrong with having it as a dependency. It's shipped virtually everywhere - it's even included in the OSGeo4W windows distribution.

kmuehlbauer:
As already written, I'm convinced :grin: Waiting for #381...

Issue #348, opened 2019-06-06
"from module import function" not recommended


egouden:
It is generally recommended not to use "from module import function" to avoid circular imports.
This decreases readability too.

kmuehlbauer:
@egouden Thanks for bringing this up!

Seems this is use throughout the project. So what is the suggestion then?

Instead of doing
```
from .projection import proj4_to_osr, reproject
```
we would do
```
from . import projection
```
and prepend `proj4_to_osr` and `reproject` with the module name (`projection`)

kmuehlbauer:
Doing a quick search through the code I found these different usages for wradlib-modules:

importing modules
- `import wradlib.ipol as ipol` 
- `from . import trafo as trafo`  (same as `from . import trafo`)

Importing functions from modules
- `from .trafo import decibel`
- `from .io.xarray import create_xarray_dataarray`
- `from . util import import_optional`

For the rewrite of the import sections I would suggest to use relative imports:
- `from . import module`

Functions/Names should not be imported directly, the module should be imported instead.

There are some locations where we already have the import statement within the code to avoid circular dependencies, see https://github.com/wradlib/wradlib/blob/396235eeadf37dc261e37575bd93be2c762dce82/wradlib/util.py#L317 .  

This is open for discussion!

@egouden Can you imagine a more fitting issue title?

egouden:
I would suggest using only import "wradlib.package as package" since our project is not that big.

kmuehlbauer:
OK! If I got your suggestion right: Since we do not have a deeply nested package is would be more straightforward to go this way:

`import wradlib.clutter as clutter`
`import wradlib.io.dem as io.dem`

egouden:
Not sure about how to deal with subpackage...

kmuehlbauer:
@egouden Can we apply from this doc: https://docs.python.org/3/tutorial/modules.html#packages ?

egouden:
Well this is the general documentation but I cannot find any recommendations regarding our issue.
Could you elaborate?
What I suggested above is what I think is best after a web search on the topic including discussions from the main python developers.
It is also recommended [here](https://docs.python-guide.org/writing/structure/).
I tend to prefer explicit imports of wradlib modules for readability and sustainability of the code but I am OK with relative imports.

Issue #341, opened 2019-06-03
extract raster dataset


egouden:
By default the function extract_raster_dataset returns pixel edge coordinates with shape (nrows+1,ncols+1,2).
This is inconsistent with the documentation.
What about the function returning pixel center coordinates as would many users expect and add a boolean parameter "edge" for this particular usage?

kmuehlbauer:
@egouden Agreed! This can be done like here:  https://docs.wradlib.org/en/stable/generated/wradlib.georef.raster.read_gdal_coordinates.html#wradlib.georef.raster.read_gdal_coordinates. PR welcome.

egouden:
PR in preparation with several improvements
I need some gdal compatible radar data for testing.
See my PR in wradlib-data.

kmuehlbauer:
@egouden Just merged the belgian radar data over in wradlib-data!

kmuehlbauer:
@egouden Let's discuss the issue here to not clutter the PR #376 with possibly unrelated stuff.

I went trough the code, again. 

Here are my findings/suggestions for further discussion:

General
----------
- [pixel_to_map3d](https://github.com/wradlib/wradlib/blob/847a3c583081747394d0e55494506b815c5af2e3/wradlib/georef/raster.py#L97) was intended to fetch the srtm data (at least, there is this note in the docstring), it will be deprecated and removed from the API, `get_raster_elevation` (see #376) will be introduced for that functionality.
- [extract_raster_dataset](https://github.com/wradlib/wradlib/blob/847a3c583081747394d0e55494506b815c5af2e3/wradlib/georef/raster.py#L309) should be the one and only API function for raster dataset extraction. It should be extended via keywords to return data, coords and proj as requested. 
[...]

PR #376, opened 2019-08-08
ENH: georef.raster 

egouden:
Here are a few improvements to the raster module.
But there are some issues:

- I had to add some imports in zonalstats to avoid circular dependencies. This is related to  #348 .
- The main change is to get center pixel coordinates instead of edge by default. This is now consistent with the doc but not backward compatible. Is this acceptable? Solution would be to keep edge by default.
- This change had impacts (failed tests) on other functions in the raster module. I tried to correct but I don't understand everything and I suspect there were some bugs. I found no impacts on other modules but I guess notebooks should be checked.
- The new function to get elevation requires downloading SRTM data which takes time for the test.
- I think some lengthy functions in the georef module could benefit from the inputs.

We should realise that this raster module has a much bigger scope than the radar community.
I am curious how many users already use this for other applications.
Therefore I think it is important to share this with other developers and put more efforts on it if needed.

BR

### DIFF
--- a/wradlib/georef/raster.py
+++ b/wradlib/georef/raster.py
@@ -22,7 +22,8 @@ Raster Functions
 import numpy as np
 from osgeo import gdal, osr, gdal_array
 
-from wradlib import georef, util
+import wradlib
+from wradlib import georef
 
 
 def _pixel_coordinates(nx, ny, mode):
@@ -266,10 +267,10 @@ def get_raster_extent(dataset, geo=False, window=True):
         y = extent[:, 1]
         extent = np.array([x.min(), x.max(), y.min(), y.max()])
 
-    return(extent)
+    return extent
 
 
-def get_raster_elevation(dataset, **kwargs):
+def get_raster_elevation(dataset, resample=None, **kwargs):
     """Return surface elevation corresponding to raster dataset
        The resampling algorithm is chosen based on scale ratio
 
@@ -277,6 +278,11 @@ def get_raster_elevation(dataset, **kwargs):
     ----------
     dataset : gdal.Dataset
         raster image with georeferencing (GeoTransform at least)
+    resample : GDALResampleAlg
+        If None the best algorithm is chosen based on scales.
+        GRA_NearestNeighbour = 0, GRA_Bilinear = 1, GRA_Cubic = 2,
+        GRA_CubicSpline = 3, GRA_Lanczos = 4, GRA_Average = 5, GRA_Mode = 6,
+        GRA_Max = 8, GRA_Min = 9, GRA_Med = 10, GRA_Q1 = 11, GRA_Q3 = 12
     kwargs : keyword arguments
         passed to wradlib.io.dem.get_strm()
 
@@ -291,23 +297,25 @@ def get_raster_elevation(dataset, **kwargs):
     driver = gdal.GetDriverByName('MEM')
     dst_ds = driver.CreateCopy('ds', dataset)
 
-    src_gt = src_ds.GetGeoTransform()
-    dst_gt = src_ds.GetGeoTransform()
-    src_scale = min(abs(src_gt[1]), abs(src_gt[5]))
-    dst_scale = min(abs(dst_gt[1]), abs(dst_gt[5]))
-    ratio = dst_scale/src_scale
-    resample = gdal.GRA_Bilinear
-    if ratio > 2:
-        resample = gdal.GRA_Average
-    if ratio < 0.5:
-        resample = gdal.GRA_NearestNeighbour
+    if resample is None:
+        src_gt = src_ds.GetGeoTransform()
+        dst_gt = dst_ds.GetGeoTransform()
+        src_scale = min(abs(src_gt[1]), abs(src_gt[5]))
+        dst_scale = min(abs(dst_gt[1]), abs(dst_gt[5]))
+        ratio = dst_scale/src_scale
+
+        resample = gdal.GRA_Bilinear
+        if ratio > 2:
+            resample = gdal.GRA_Average
+        if ratio < 0.5:
+            resample = gdal.GRA_NearestNeighbour
 
     gdal.ReprojectImage(src_ds, dst_ds,
                         src_ds.GetProjection(), dst_ds.GetProjection(),
                         resample)
     elevation = read_gdal_values(dst_ds)
 
-    return(elevation)
+    return elevation
 
 
 def set_raster_origin(data, coords, direction):
@@ -553,7 +561,7 @@ def merge_rasters(datasets):
 
     dataset = gdal.Warp('', datasets, format='MEM')
 
-    return(dataset)
+    return dataset
 
 
 def raster_to_polyvert(dataset):
@@ -572,6 +580,6 @@ def raster_to_polyvert(dataset):
     """
     rastercoords = read_gdal_coordinates(dataset, mode="edges")
 
-    polyvert = util.grid_to_polyvert(rastercoords)
+    polyvert = georef.rect.grid_to_polyvert(rastercoords)
 
-    return(polyvert)
+    return polyvert
--- a/wradlib/georef/raster.py
+++ b/wradlib/georef/raster.py
@@ -36,21 +36,21 @@ def _pixel_coordinates(nx, ny, mode):
     ny : int
         y size (numbers or rows)
     mode : string
-        either 'centers' (0.5 1.5 ...) or 'edges' (0 1 ...)
+        either 'center' (0.5 1.5 ...) or 'edge' (0 1 ...)
 
     Returns
     -------
     coordinates : :class:`numpy:numpy.ndarray`
          array containing pixel coordinates (x,y) in image convention
-         shape is (nrows, ncols, 2) if mode==centers
-         shape is (nrows+1, ncols+1, 2) if mode==edges
+         shape is (nrows, ncols, 2) if mode==center
+         shape is (nrows+1, ncols+1, 2) if mode==edge
 
     """
-    if mode == "centers":
+    if mode == "center":
         x = np.linspace(0.5, nx-0.5, num=nx)
         y = np.linspace(0.5, ny-0.5, num=ny)
 
-    if mode == "edges":
+    if mode == "edge":
         x = np.linspace(0, nx, num=nx + 1)
         y = np.linspace(0, ny, num=ny + 1)
 
@@ -93,7 +93,7 @@ def _pixel_to_map(coordinates, geotransform):
     return coordinates_map
 
 
-def read_gdal_coordinates(dataset, mode="centers"):
+def read_gdal_coordinates(dataset, mode="center"):
     """Get the projected coordinates from a GDAL dataset.
 
     Parameters
@@ -101,7 +101,7 @@ def read_gdal_coordinates(dataset, mode="centers"):
     dataset : gdal.Dataset
         raster image with georeferencing
     mode : string
-        either 'centers' or 'edges'
+        either 'center' or 'edge'
 
     Returns
     -------
@@ -109,7 +109,7 @@ def read_gdal_coordinates(dataset, mode="centers"):
         Array of shape (nrows,ncols,2) containing xy coordinates.
         The array indexing follows image convention with origin
         at upper left pixel.
-        The shape is (nrows+1,ncols+1,2) if mode == edges.
+        The shape is (nrows+1,ncols+1,2) if mode == edge.
 
     Examples
     --------
@@ -190,7 +190,7 @@ def read_gdal_values(dataset=None, nodata=None):
     return np.squeeze(np.array(bands))
 
 
-def extract_raster_dataset(dataset, mode="centers", nodata=None):
+def extract_raster_dataset(dataset, mode="center", nodata=None):
     """ Extract data, coordinates and projection information
 
     Parameters
@@ -198,7 +198,7 @@ def extract_raster_dataset(dataset, mode="centers", nodata=None):
     dataset : gdal.Dataset
         raster dataset
     mode : string
-        either 'centers' or 'edges'
+        either 'center' or 'edge'
     nodata : float
         replace nodata values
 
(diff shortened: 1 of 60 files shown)


wradlib/georef/raster.py
```diff
@@ -27,37 +27,40 @@
 from osgeo import gdal, osr, gdal_array
 
 from .projection import reproject
+from ..io import dem
 
 
-def pixel_coordinates(nx, ny, mode="centers"):
-    """Get pixel coordinates from a regular grid with dimension nx by ny.
+def pixel_coordinates(nx, ny, edge=False):
```
kmuehlbauer: We can't just switch to the new `edge` kwarg since this would introduce a major version bump according to `semver`. We would need to have both kwargs side by side and issue a deprecation warning when using the old one.

wradlib/georef/raster.py
```diff
@@ -27,37 +27,40 @@
...
     Parameters
     ----------
     nx : int
         xsize
     ny : int
         ysize
-    mode : string
-        `centers` or `edges` to return the pixel coordinates
-        defaults to centers
+    edge : bool
```
kmuehlbauer: Might just leave the string kwarg as it is.

wradlib/georef/raster.py
```diff
@@ -256,7 +256,9 @@ def create_raster_dataset(data, coords, projection=None, nodata=-9999):
 
     # initialize geotransform
     x_ps, y_ps = coords[1, 1] - coords[0, 0]
-    geotran = [coords[0, 0, 0], x_ps, 0, coords[0, 0, 1], 0, y_ps]
+    upper_corner_x = coords[0, 0, 0] - x_ps/2
```
kmuehlbauer: This is only true, if we have "center" coordinates. The function was implemented by using the UPPER LEFT pixel corner as this is also the anchor point for the gdal geotransform.

wradlib/georef/raster.py
```diff
@@ -475,3 +479,87 @@ def reproject_raster_dataset(src_ds, **kwargs):
...
+        corner coordinates [ul,ll,lr,ur] or
+        window extent [xmin, xmax, ymin, ymax]
+    """
+
+    x_size = dataset.RasterXSize
+    y_size = dataset.RasterYSize
+    geotrans = dataset.GetGeoTransform()
+    xmin = geotrans[0]
+    ymax = geotrans[3]
+    xmax = geotrans[0] + geotrans[1] * x_size
```
kmuehlbauer: This covers the UPPER LEFT corner but not the LOWER RIGHT corner.

wradlib/georef/raster.py
```diff
@@ -300,13 +302,13 @@ def set_raster_origin(data, coords, direction):
         coords = np.flip(coords, axis=-3)
         # we need to shift y-coordinate if data and coordinates have the same
         # number of rows and cols (only the ll or ul raster coords are given)
-        if data.shape[-2:] == coords.shape[:2]:
-            coords += [0, y_sp]
+#        if data.shape[-2:] == coords.shape[:2]:
```
kmuehlbauer: Here too, pixel corner coordinates are assumed. If we only have the upper left coordinates of the pixels we need to shift the coordinates after flipping because upper left will then be lower left.

wradlib/georef/raster.py
egouden: I do not understand.

kmuehlbauer: If geotrans[0]/geotrans[3] is the UPPER LEFT corner point, the the LOWER RIGHT corner point would be

geotrans[0] + getrans[1] * x_size + x_size
geotrans[3] + getrans[3] * y_size + y_size

wradlib/georef/raster.py
egouden: I could keep the current API.
But what is your general opinion on API change and version in wradlib?

wradlib/georef/raster.py
egouden: In this case center coordinates are easier ;-)

egouden:
Thank you for considering the changes.
The prime motivation for these changes was #341: currently in part of the API center coordinates are found but edges coordinates are used instead. I also think that center coordinates are more user friendly.
I see some advantages of using edges in the implementation of some functions in relation to GDAL geotransform but I don't see a clear guidance from the GDAL doc.
So I think we should at least fix the API documentation and I let you decide which implementation you prefer.
I am ready to implement your decision.

wradlib/georef/raster.py
egouden: Yes I know you used the edge coordinates.
This is just the center coordinates version, nothing wrong with your version.

wradlib/georef/raster.py
egouden: Yes indeed

kmuehlbauer:
@egouden Thanks for the comments. I think it would be good if we list the usages within wradlib and within known users workflows (yours, mine etc.). Then we might get a better insight, how we should finalize this.

wradlib/georef/raster.py
kmuehlbauer: - enhancements and non-breaking changes -> increase minor version 
- breaking changes -> increase major version

If we think an API change makes really sense, we can issue deprecation warnings for the old API while accepting old and new API. After some minor version bumps, we can deprecate the old API and release a new major version.

wradlib/georef/raster.py
egouden: In a four pixels image with size 1 and UL corner at 0,0 LR corner would be:
0 + 1 * 2 = 2
0 - 1 * 2 = -2
My tests with actual values are also successful.
Am I missing something?

kmuehlbauer: @egouden Yes, you are right and I'm wrong. Not sure what I was thinking when I wrote it. Sorry for the confusion...

kmuehlbauer:
@egouden The import stuff needs revision. I'll do that in another PR.

egouden:
A concrete suggestion for the imports and namespaces with readability in mind...

kmuehlbauer:
@egouden `imports` are breaking on Python 3.6 most likely due to importing uninitialized submodule/package.

My suggestion would be to use:

```python
from wradlib import georef, io, util
```
or

```python
from wradlib import georef as georef
from wradlib import io as io
from wradlib import util as util
```

From my perspective the former looks nicer.

If we restrict us to import only submodules and no functions, there should be no problem.

wradlib/georef/raster.py
```diff
@@ -14,58 +14,61 @@
...
    set_raster_origin
-   extract_raster_dataset
 """
 import numpy as np
 from osgeo import gdal, osr, gdal_array
 
-from wradlib.georef import projection
+import wradlib
+import wradlib.georef as georef
+import wradlib.util as util
```
kmuehlbauer: See https://github.com/wradlib/wradlib/pull/376#issuecomment-530695386

wradlib/georef/raster.py
```diff
@@ -217,57 +191,125 @@ def read_gdal_values(dataset=None, nodata=None):
...
-    if projection:
-        dataset.SetProjection(projection.ExportToWkt())
+    if window:
+        x = extent[:, 0]
+        y = extent[:, 1]
+        extent = np.array([x.min(), x.max(), y.min(), y.max()])
 
-    # set np.nan to nodata
-    dataset.GetRasterBand(1).SetNoDataValue(nodata)
+    return(extent)
```
kmuehlbauer: Please do not use braces in return statement.

wradlib/georef/raster.py
```diff
@@ -217,57 +191,125 @@ def read_gdal_values(dataset=None, nodata=None):
...
+
+    driver = gdal.GetDriverByName('MEM')
+    dst_ds = driver.CreateCopy('ds', dataset)
+
+    src_gt = src_ds.GetGeoTransform()
+    dst_gt = src_ds.GetGeoTransform()
+    src_scale = min(abs(src_gt[1]), abs(src_gt[5]))
+    dst_scale = min(abs(dst_gt[1]), abs(dst_gt[5]))
+    ratio = dst_scale/src_scale
+    resample = gdal.GRA_Bilinear
```
kmuehlbauer: Would be nice to make `resample` a kwarg which overrides the built-in behaviour.

wradlib/georef/raster.py
kmuehlbauer: Shifting is not needed any more, since we have now proper alignment . Just delete the remnants.

wradlib/georef/raster.py
```diff
@@ -217,57 +191,125 @@ def read_gdal_values(dataset=None, nodata=None):
...
+        resample = gdal.GRA_Average
+    if ratio < 0.5:
+        resample = gdal.GRA_NearestNeighbour
+
+    gdal.ReprojectImage(src_ds, dst_ds,
+                        src_ds.GetProjection(), dst_ds.GetProjection(),
+                        resample)
+    elevation = read_gdal_values(dst_ds)
+
+    return(elevation)
```
kmuehlbauer: remove braces

wradlib/georef/raster.py
```diff
@@ -474,3 +482,98 @@ def reproject_raster_dataset(src_ds, **kwargs):
...
+    bands, rows, cols = data.shape
+
+    # create In-Memory Raster with correct dtype
+    mem_drv = gdal.GetDriverByName('MEM')
+    gdal_type = gdal_array.NumericTypeCodeToGDALTypeCode(data.dtype)
+    dataset = mem_drv.Create('', cols, rows, bands, gdal_type)
+
+    # initialize geotransform
+    x_ps, y_ps = coords[1, 1] - coords[0, 0]
+    upper_corner_x = coords[0, 0, 0] - x_ps/2
```
kmuehlbauer: This assumes `centers`, right? IIRC, geotransform uses the corners of the pixels not the centers.

wradlib/georef/raster.py
```diff
@@ -474,3 +482,98 @@ def reproject_raster_dataset(src_ds, **kwargs):
...
+    # set np.nan to nodata
+    dataset.GetRasterBand(1).SetNoDataValue(nodata)
+
+    for i, band in enumerate(data, start=1):
+        dataset.GetRasterBand(i).WriteArray(band)
+
+    return dataset
+
+
+def merge_rasters(datasets):
```
kmuehlbauer: Just a nit: would `merge_raster_datasets` be a longer but more fitting name?

wradlib/georef/raster.py
```diff
@@ -474,3 +482,98 @@ def reproject_raster_dataset(src_ds, **kwargs):
...
+    datasets : list of gdal.Dataset
+        raster images with georeferencing
+
+    Returns
+    -------
+    dataset : gdal.Dataset
+        merged raster dataset
+    """
+
+    dataset = gdal.Warp('', datasets, format='MEM')
```
kmuehlbauer: Would be nice, if we could drop a `gdal.WarpOptions()` as kwargs.

wradlib/georef/raster.py
```diff
@@ -474,3 +482,98 @@ def reproject_raster_dataset(src_ds, **kwargs):
...
+
+    return(dataset)
```
kmuehlbauer: no braces

wradlib/georef/raster.py
```diff
@@ -474,3 +482,98 @@ def reproject_raster_dataset(src_ds, **kwargs):
...
+
+    Parameters
+    ----------
+    dataset : gdal.Dataset
+        raster image with georeferencing (GeoTransform at least)
+
+    Returns
+    -------
+    polyvert : :class:`numpy:numpy.ndarray`
+        A 3-d array of polygon vertices with shape (..., 5, 2).
```
kmuehlbauer: `3-d ` is somehow misleading. Do we get (K, M, N, 5, 2)? The it would be 5-d, Or is this arbitrary, then we could state `N-d`.

wradlib/georef/raster.py
```diff
@@ -474,3 +482,98 @@ def reproject_raster_dataset(src_ds, **kwargs):
...
+
+    """
+    rastercoords = read_gdal_coordinates(dataset, mode="edges")
+
+    polyvert = util.grid_to_polyvert(rastercoords)
```
kmuehlbauer: I would rather see the function `grid_to_polyvert` in the `wradlib.georef.rect` submodule.

wradlib/georef/raster.py
```diff
@@ -474,3 +482,98 @@ def reproject_raster_dataset(src_ds, **kwargs):
...
+
+    return(polyvert)
```
kmuehlbauer: no braces

wradlib/tests/test_georef.py
```diff
@@ -2,13 +2,17 @@
...
 import unittest
 
 import numpy as np
 from osgeo import gdal, osr, ogr
 import xarray as xr
 
-from wradlib import georef, io, util
+import wradlib
+import wradlib.georef as georef
+import wradlib.util as util
```
kmuehlbauer: see https://github.com/wradlib/wradlib/pull/376#issuecomment-530695386

wradlib/tests/test_georef.py
```diff
@@ -2,13 +2,17 @@
 # Copyright (c) 2011-2019, wradlib developers.
 # Distributed under the MIT License. See LICENSE.txt for more info.
 
+import sys
+
```
kmuehlbauer: remove empty line

wradlib/georef/raster.py
kmuehlbauer: OK got it, we are assuming here `coords` are center-coords, not edge coords. Should we make this clear somewhere? And, what if the user already has edge coordinates?

wradlib/util.py
```diff
@@ -639,5 +639,33 @@ def gradient_from_smoothed(x, n=5):
...
 
 
+def grid_to_polyvert(grid, ravel=False):
+    """Get polygonal vertices from rectangular grid coordinates.
+
+    Parameters
+    ----------
+    grid : numpy array
+        grid edge coordinates
+
```
kmuehlbauer: `ravel` missing

wradlib/tests/test_georef.py
kmuehlbauer: below `import sys`

kmuehlbauer:
@egouden Is this ready to go from your end? Minor things could be fixed later, I think.

egouden:
It is OK for me now but could you check again?
I think this will be popular.

kmuehlbauer:
@egouden Thanks for this valuable contribution and for your patience and endurance in the review process!

PR #379, opened 2019-09-10
improvement projection

egouden:
This adds a few projection tools including geoid calculations.

### DIFF
--- a/wradlib/georef/projection.py
+++ b/wradlib/georef/projection.py
@@ -14,7 +14,9 @@ Projection Functions
    {}
 """
 __all__ = ['reproject', 'create_osr', 'proj4_to_osr', 'epsg_to_osr',
-           'wkt_to_osr', 'get_default_projection', 'get_earth_radius']
+           'wkt_to_osr', 'get_default_projection', 'get_earth_radius',
+           'get_radar_projection', 'get_earth_projection',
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
@egouden This is rebased to latest master, please checkout your local branch correctly. I've had problems to get one test to run. Also did not review, yet. There might be room for further improvement.

kmuehlbauer:
Seems that your PR depend on each other `raster->ipol->projection->polar`. I tried to rebase everything but this might not be the best approach. So I'll wait for your turn.

egouden:
This PR is independent.
Note that GDAL >= 2.2 is required for vertical datum shift.
It still fails on windows.
Maybe because of this file missing 
Since I have no access to Windows machines, could you investigate this issue?

kmuehlbauer:
@egouden I've reassigned `strict_channel_priority` for appveyor. Let's see if things work now better.

kmuehlbauer:
@egouden Hmm, although we got a newer gdal it still breaks. Can you just mark it as `skip` on windows?

egouden:
Well shouldn't this work also on windows?

kmuehlbauer:
@egouden Yes it should, but for some reason it doesn't. Until we figure out why it doesn't work, we can skip the test.

kmuehlbauer:
@egouden Thanks! Nice addition, but we need to reiterate a bit before release of wradlib 1.7.

PR #380, opened 2019-09-10
Improvement/polar

egouden:
This is my take on georeferencing. 
It includes optimization of existing features and some new features with, in particular:
- accurate computation of sweep map coordinates based on the sphere model (I think the current default WGS84 radar projection is not consistent with the model)
- a reverse transformation from map coordinates to radar coordinates
- a function to convert a raster to a sweep and vice versa


kmuehlbauer:
@egouden I'll move this to the next milestone.

kmuehlbauer:
@egouden Too much has changed over the time. It would be great to have your additions merged. I suggest to update the PR or come up with a completely new PR. I'll close for now. Please reopen after updating that PR. Thanks!

PR #381, opened 2019-09-10
MNT: fix imports, license header, requirements

### LINKED ISSUE #348: "from module import function" not recommended

### LINKED ISSUE #375: nbformat and nbconvert are needed for testrunnery.py but not in the requirements

kmuehlbauer:
fixes #348
fixes #375

The imports in wradlib modules follow now the [PEP8 recommendation](https://pep8.org/#imports):
eg.

```python
import os
import sys

import numpy as np
import xarray as xr

from wradlib import io, ipol
from wradlib.georef import projection
```

Absolute imports are used throughout the package.

We would have some name clash with system and wradlib `io` module. This is circumvented using ìmport io as sio`. Other possible name clashes with eg. `gdal` are possible but can also be mitigated.

The function `maximum_intensity_projection` was moved from `util` to `georef.polar` because it uses polar data and the circular dependency is avoided. There is only one circular dependency left in `io.hdf` (import zonalstats.get_clip_mask). This can also be removed after some refactoring in a later PR.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2011-2018 wradlib developers
+Copyright (c) 2011-2019 wradlib developers
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,9 @@
-numpy
-scipy
-matplotlib
+deprecation
+gdal
 h5py
 netCDF4
+numpy
+matplotlib
+scipy
 xarray
-gdal
-deprecation
 xmltodict
-semver
--- a/requirements_devel.txt
+++ b/requirements_devel.txt
@@ -1,4 +1,8 @@
-coverage
 codecov
+coverage
 pytest
 pytest-cov
+pytest-sugar
+pytest-xdist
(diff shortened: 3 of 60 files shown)


egouden:
Why is the diff showing some files completely changed?

kmuehlbauer:
@egouden probably line ending or whitespace issues.

kmuehlbauer:
@egouden It's whitespace changes. You can disable this while reviewing. There is a small *setup wheel* in the upper left area of the review page, where you can disable this.

kmuehlbauer:
@egouden If you agree with this, I'll merge next day to go ahead with your pending PR's.

kmuehlbauer:
Update: added handling of moved/deprecated functions

kmuehlbauer:
@egouden This is now fixed. I'll merge right away, if you approve.

egouden:
Nice job the imports are now consistent.
I am OK with the changes to move on but I still have a few remarks.

I would only mention PEP8 as a starting point.
This is old (2001) and as the general documentation do not actually provides recommendations to solve our issues.

Circular dependencies could not always be avoided but should not be an issue with the proposed change. This is the whole point of #348.  Do you really get an error?

I would prefer using only georef to access all georefencing functions because the subpackages structure is a bit subjective in my opinion and add unnecessary complexity to the code. 

I would also not use the "from" at all as discussed earlier.

kmuehlbauer:
@egouden For `georef` and `io` modules, all functions from subpackages are imported into the respective module via module __init__.py.

So for the user all functions are available via `wradlib.georef.function` or `wradlib.io.function`. 

Subpackage structure might be subjective but is necessary to keep the code readable. It went a long way to the current layout but there might still be room for consolidation and improvement.

PR #383, opened 2019-09-15
Correct error in documentation of sweep_centroids

ElmerJeanpierreLopez:
Documentation mentions elevation angle in [rads] ,but it should be [deg]. I already discussed in Issue #9.


kmuehlbauer:
Thanks @ElmerJeanpierreLopez!

Issue #373, opened 2019-08-03
Interpolation kwarg not passed to interpolator call method


jorahu:
Using Nearest interpolator with maxdist kwarg when using wradlib.comp.togrid function does not work although togrid documentation says kwargs are passed to interpolator class. 
https://github.com/wradlib/wradlib/blob/d7e30e61bcdb6faed122e8b41926c88ef1650ed4/wradlib/comp.py#L89 calls the interpolator without maxdist kwarg. 
When calling the comp.togrid method with maxdist gives TypeError: TypeError: __init__() got an unexpected keyword argument 'maxdist'.

kmuehlbauer:
@jorahu Sorry, this had to be postponed to the next (bugfix-) release. It's a bit more complicated thatn anticipated and might need an rewrite of the interpolator classes.

kmuehlbauer:
@jorahu In #425 I've implemented a bit of a hack to get this in. You can add the calling keyword arguments by adding
```python
togrid(...,...,..., call_kwargs=dict(maxdist=10))
```
This will be available in wradlib 1.7 in about two weeks time.

PR #387, opened 2019-09-17
improvement ipol

egouden:


kmuehlbauer:
@egouden Just curious, could you elaborate a bit on your extension to scipy's `binned_statistic_dd`? If this is some standard use case, we should try to implement this in `scipy` directly.

egouden:
@kmuehlbauer 
The idea is to reuse the bins for saving time.
It is certainly a standard use case.
Do you think contributing to scipy is easy?

kmuehlbauer:
> The idea is to reuse the bins for saving time.

Makes sense, so you call it once `values=None` and get the `binnumbers` back for use in a forthcoming call?

> It is certainly a standard use case.

Then it should be implemented within scipy.

> Do you think contributing to scipy is easy?

Not sure, but you could open an issue explaining your use and outlining your solution at the scipy github issue tracker.

kmuehlbauer:
@egouden While you are on it, do you see an easy solution for #373?

egouden:
@kmuehlbauer 
You just need to pass maxdist to the interpolator call in comp.py
I would not allow maxdist to be initialised in ipol.py

egouden:
> > The idea is to reuse the bins for saving time.
> 
> Makes sense, so you call it once `values=None` and get the `binnumbers` back for use in a forthcoming call?
> 
> > It is certainly a standard use case.
> 
> Then it should be implemented within scipy.
> 
> > Do you think contributing to scipy is easy?
> 
> Not sure, but you could open an issue explaining your use and outlining your solution at the scipy github issue tracker.

I just send a message to the scipy-dev list.

Issue #386, opened 2019-09-16
georef init


egouden:
Why not using from module import * in the georef init?
The current situation is really time consuming when developing.

kmuehlbauer:
I find it easier to maintain a stable API (in terms of functions). Another point if we would do this (eg `from .polar import *`) all other packages imported within the subpackage will clutter the namespace (like numpy, warnings etc).

egouden:
OK.
What about keeping only the most useful functions?

kmuehlbauer:
If we identify functions which should not be used via `wradlib.georef.function` (but can still be reached via `wradlib.georef.subpackage.function`) we might remove them from `georef`. 

Which do you have in mind?

egouden:
I think my last suggestion makes things complicated.
In fact as already mentioned I prefer not to use the georef submodules.
Therefore I am fine with the current situation.
I just hope there is a way to automatically import only the functions.

kmuehlbauer:
@egouden Yes, I'm still looking for a solution to this. If you fiond anything by chance, let me know.

kmuehlbauer:
@egouden I have an idea for a solution:

In the submodule we do something like this:

```python
"""
Miscellaneous
^^^^^^^^^^^^^

.. autosummary::
   :nosignatures:
   :toctree: generated/

   {}
"""
__all__ = ['bin_altitude', 'bin_distance', 'site_distance']
__doc__ = __doc__.format('\n   '.join(__all__))
```

We have to specify the functions which should be exported/documented in the `__all__` variable of the respective submodule. This will then also be added to the docstring.

In the module we would then be able to use `from .misc import *` etc..

I tested this and can issue a PR if you think this will be more easy to use.

kmuehlbauer:
resolved by #388

PR #388, opened 2019-09-17
MNT: structure submodule imports

kmuehlbauer:
use `__all__` in submodules (georef, io) to specify exported/documented functions


kmuehlbauer:
@egouden possible solution for #386

kmuehlbauer:
Looks like adding `__all__` in every module is generally a good idea.

kmuehlbauer:
@egouden Could you have a look at this? I'd like to pull this in near time.

kmuehlbauer:
@egouden At least we'll have only a single point of failure. :grin:

PR #392, opened 2019-09-26
Improvement/multixarray

egouden:
This allows to read successive sweeps by concatenation over azimuth or merging over time.
Currently the sweeps have to be consistent for a given elevation angle.

### DIFF
--- a/wradlib/io/xarray.py
+++ b/wradlib/io/xarray.py
@@ -1142,9 +1142,11 @@ class OdimH5(XRadVol):
         self.standard = kwargs.get('standard', 'cf-mandatory')
         self.dim0 = kwargs.get('dim0', 'azimuth')
 
-        self.ds = None
+        self.ds = []
+        self._nch = []
         for f in filename:
             nch = OdimH5File(f, flavour=flavour)
+            self._nch.append(nch)
 
             # retrieve and assign global groups /how, /what, /where
             groups = ['how', 'what', 'where']
@@ -1190,7 +1192,7 @@ class OdimH5(XRadVol):
                 print("last variable")
                 print(i)
                 allvars.append(ds)
-                allvars = xr.merge(allvars)
+                allvars = xr.merge(allvars, compat='override')
                 merged_ds.append(allvars)
                 allvars = []
         self.ds = merged_ds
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
@egouden I like the idea to assign the `root` object separately after assignment of the sweeps. 
Unfortunately the current implementation breaks some code ([this notebook](https://github.com/wradlib/wradlib-notebooks/blob/1.5.0/notebooks/fileio/wradlib_load_DWD_opendata_volumes.ipynb)). I'll have a look, why and when this happens, but you might also have a look.

Another thing is for the naming of the `scan`-dimension. There are three options from my point of view :

- rename the original `time`-dimension/coordinate and use new `time` dimension instead of `scan`
- use 'volume` instead of `scan`
- keep `scan`

I would tend to use new `time`-dimension, together with `dim0='azimuth'`. This is violating CfRadial2 standard, but we should keep the implementation simple. As long as we can read and write CfRadial2 compliant files everything should be fine.

egouden:
@kmuehlbauer 
I let you check the notebook because I am not familiar with it.

I also prefer using the 'azimuth' and extra 'time' dimension.
At least this could be default.

kmuehlbauer:
@egouden Just to let you know, why this brakes: The German Weather Service distributes it's data in a so called *one sweep, one moment, one file" manner. This means one sweep consists of several files with only one moment inside. The current changes do not detect, if different moment files belong to the same sweep.

kmuehlbauer:
@egouden Another thing which need to be taken into account: What if the data layout differs between volumes, due to some failures. I've seen double rays, missing rays and the like. It would break while merging datasets, then.

kmuehlbauer:
@egouden OK, this is a hard nut to crack. For xarray to make the concatenation work correct, I would need to merge all moments of one sweep at one time **before** concatenating along the new `time` dimension (what is exactly what OdimH5 was supposed to do). Looking at the code (which is already complex enough) I do not see a nice and elegant way of doing this.

I would suggest that we leave the OdimH5 being a container for **one volume**. Instead we can invent some `VolumeTimeSeries` which can take several OdimH5/CfRadial volumes to create a timeseries.

egouden:
This make it less complex I think.
But I am again stuck with the NETCDF4 file ID issue.
It seems the files are closed under the hood when I use a list to store the datasets.

egouden:
One day of wandering to finally find the issue: keep nch handles !
This can sort and merge any sweeps using time or azimuth coordinates.

kmuehlbauer:
@egouden I took your ideas of merge/concat into #395. Referenced you as Co-Author.

PR #393, opened 2019-10-02
FIX: single dimension problem

kmuehlbauer:
1. fix single dimension problem (nrays=ndims)
2. assign root after reading sweeps (moved from @egouden)

done in #395 
3. VolumeTimeSeries class + test (VolTS, with @egouden) 

Test data for 1.& 3. possibly provided by BOM Australia, waiting for feedback

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,9 @@
 !/wradlib/
 !/wradlib/*
 
+# ignore version.py
+wradlib/version.py
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/wradlib/georef/xarray.py
+++ b/wradlib/georef/xarray.py
@@ -164,10 +164,13 @@ def georeference_dataset(obj, **kwargs):
 
     dim0 = obj['azimuth'].dims[0]
 
+    # create meshgrid to overcome dimension problem with spherical_to_xyz
+    r, az = np.meshgrid(obj['range'], obj['azimuth'])
+
     # GDAL OSR, convert to this proj
     if isinstance(proj, osr.SpatialReference):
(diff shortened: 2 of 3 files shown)


kmuehlbauer:
@egouden You might have a look into this. I had to copy your `assign_root` part due to the other changes a simple merging wasn't actually feasable. Tomorrow is a day off here, hope to get your things in on friday.

egouden:
OMG
I have been playing with this as well... see my updated PR.
Does your VolTS works from the DWD files?

egouden:
Yes we can discuss this on Tuesday ;-)
Now I am taking some holidays...

kmuehlbauer:
> Does your VolTS works from the DWD files?

Yes, it works well with the DWD data. I've checked with an updated version of the DWD notebook. Now the real adventure can start.

egouden:
It does not work with sweep by sweep inputs.

kmuehlbauer:
> What if my scanning strategy is like this:

| Volume | swp 1 | swp 2 | swp 3 |
| --- | --- | --- | --- |
| 0 | 3 | 1.6 | 1 |
| 1 | 4 | 1.6 | 1 |
| 2 | 2 | 1.6 | 1 |
| 3 | 3 | 1.6 | 1 |
| 4 | 4 | 1.6 | 1 |
| 5 | 2 | 1.6 | 1 |
| .. | .. | .. | .. | 

Right, this doesn't work. The requirement on the input files:

- OdimH5: files have different elevation angles (for PPI) and belong to same volume (eg. DWD data)
- OdimTS: files contain same structure of sweeps.

We just can't combine different elevation angles (with possibly different number of bins and gate length) into one consistent xarray dataset.

kmuehlbauer:
taken over by #395

PR #395, opened 2019-10-29
ENH: MultiXarray

kmuehlbauer:
This is based on work of @egouden and @kmuehlbauer.

Only the current classes are advertised via the docstrings. This will change for future wradlib 2.0.


kmuehlbauer:
resolved by #397

PR #396, opened 2019-10-29
ADD: add @egouden to Developers section

kmuehlbauer:


kmuehlbauer:
@egouden I'm sorry for misspelling your name. Thank you for pointing out the correct affiliation. Fixed now.

PR #399, opened 2019-12-23
MNT: use semver 2.9.0, fix setup.py

kmuehlbauer:


kmuehlbauer:
Seems that something broke conda-forge's `semver`. Using pip version for now.

PR #401, opened 2020-02-01
FIX: low cloud switch removes everything

egouden:
There is an error in the logical expression for the low switch in clutter.filter_cloudtype.
It implies that all data are identified as clutter.
Now the cloud types are considered explicitly using equality.
A test has been added to discover this bug.

### DIFF
--- a/wradlib/clutter.py
+++ b/wradlib/clutter.py
@@ -494,12 +494,12 @@ def filter_cloudtype(img, cloud, thrs=0, snow=False, low=False, cirrus=False,
     thrs : float
         Threshold above which to identify clutter
     snow : bool
-        Swith to use PGE02 class "land/sea snow" for clutter identification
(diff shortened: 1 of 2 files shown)


kmuehlbauer:
@egouden Thanks for fixing this. Can you have a look, why the new test is failing? Otherwise we can merge, if the CI is OK.

kmuehlbauer:
@egouden What is the expected `result` of the `filter_cloudtype` in the test? Obviously it returns a boolean mask of the detected clutter.

If checked with different values of `self.val` and get different results (which makes sense, right). But the assertion need to be adapated. I'd rather build on your expertise to add the correct values.

kmuehlbauer:
@egouden This might need a rebase, then it is good to go...

kmuehlbauer:
@egouden I'll merge this, when the CI is happy...

PR #407, opened 2020-02-17
enh: fix double rays before merging

egouden:
Currently radar moments with same elevation and same timestamp are merged before obtaining coordinates.
Inconsistencies in the moments files could not be detected or corrected
With this PR the coordinates are obtained and fixed before merging.

### DIFF
--- a/wradlib/io/xarray.py
+++ b/wradlib/io/xarray.py
@@ -766,19 +766,36 @@ def _preprocess_moment(ds, mom):
     ds = ds.rename({'data': quantity,
                     dims[0]: 'azimuth',
                     dims[1]: 'range'})
+    if mom.parent.decode_coords:
+        coords = mom.parent._get_coords().coords
+        ds = ds.assign_coords(coords)
+        ds = ds.sortby('azimuth')
+        ds = ds.pipe(_reindex_azimuth, mom.parent)
 
     return ds
 
 
 def _reindex_azimuth(ds, sweep, force=False):
+
     dim = list(ds.dims)[0]
     diff = ds[dim].diff(dim)
+
     # this captures missing/additional rays and different angle spacing
     if force | (len(set(diff.values)) > 1):
+        warnings.warn("WRADLIB: missing/additional rays detected "
+                      "=> trying to fix", UserWarning)
+
         # find exact duplicates and remove
         _, idx = np.unique(ds['azimuth'], return_index=True)
         if len(idx) < len(ds['azimuth']):
             ds = ds.isel(azimuth=idx)
+
+            # recompute rtimes if based on nrays
+            sweep._nrays = len(idx)
+            sweep._rtime = None
+            if len(sweep.ray_times) == len(idx):
+                ds = ds.assign({'rtime': sweep.ray_times})
+
         # create new array and reindex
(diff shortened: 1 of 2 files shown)


wradlib/io/xarray.py
```diff
@@ -1691,6 +1706,12 @@ def _merge_moments(self):
             ds = (ds.assign_coords({'azimuth': self.azimuth}).sortby('azimuth')
                   .pipe(_reindex_azimuth, self)
                   .drop('azimuth'))
+        else:
+            coords = mom.parent._get_coords().coords
```
kmuehlbauer: This is `self` instead of `mom.parent` since GAMIC has all moments already in one dataset.

wradlib/io/xarray.py
```diff
@@ -1691,6 +1706,12 @@ def _merge_moments(self):
...
+            ds = ds.assign_coords(coords)
+            ds = ds.sortby('azimuth')
+            ds = ds.pipe(_reindex_azimuth, mom.parent)
```
kmuehlbauer: Same here (`self` instead of `mom.parent`)

wradlib/io/xarray.py
```diff
@@ -1411,10 +1425,11 @@ def _get_azimuth_how(self):
         zero_index = np.where(stopaz < startaz)
         stopaz[zero_index[0]] += 360
         azimuth_data = (startaz + stopaz) / 2.
+        azimuth_data = np.round(azimuth_data, decimals=1)
```
kmuehlbauer: We would need to do the rounding also in the GAMIC related function....

wradlib/io/xarray.py
```diff
@@ -766,19 +766,36 @@ def _preprocess_moment(ds, mom):
...
 
 
 def _reindex_azimuth(ds, sweep, force=False):
+
     dim = list(ds.dims)[0]
     diff = ds[dim].diff(dim)
+
     # this captures missing/additional rays and different angle spacing
     if force | (len(set(diff.values)) > 1):
+        warnings.warn("WRADLIB: missing/additional rays detected "
```
kmuehlbauer: Can we silence this warning, after some occurrences? If we load a timeseries of 288 sweeps, and half of the are affected, this will clutter the log...

wradlib/io/xarray.py
```diff
@@ -766,19 +766,36 @@ def _preprocess_moment(ds, mom):
...
+                      "=> trying to fix", UserWarning)
+
         # find exact duplicates and remove
         _, idx = np.unique(ds['azimuth'], return_index=True)
         if len(idx) < len(ds['azimuth']):
             ds = ds.isel(azimuth=idx)
+
+            # recompute rtimes if based on nrays
```
kmuehlbauer: can you explain, what this is supposed to do?

wradlib/io/xarray.py
```diff
@@ -766,19 +766,36 @@ def _preprocess_moment(ds, mom):
...
+            sweep._nrays = len(idx)
```
kmuehlbauer: Ahh, I see, you are cheating/overriding the nrays/and rtime to be recalculated for the correct amount of rays.

PR #419, opened 2020-04-24
ENH: improve derivation function

kmuehlbauer:
- break out derivation code to `util`
- improve derivation itself
- improve NaN - handling

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,8 +3,8 @@ environment:
   CONDA_PATH: "C:\\Miniconda37-x64"
 
   matrix:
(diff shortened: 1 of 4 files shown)


kmuehlbauer:
This speeds up KDP derivation (with NaN handling) by 2 to 3 orders of magnitude.

So if you consider a sweep (az: 360, range: 1000) with equally spaced NaN along the range every 10 gates. If you further consider a window length of 7, then the processing time drops from around 1 minute to approx. 100 ms. Your mileage may vary.

PR #421, opened 2020-04-30
WIP: ENH: Speedup unfold phi vulpiani

kmuehlbauer:
Slight improvement in kdp based phase unwrapping:

- takes now parameter th and winlen
- unfolds first phase wrap
- fixes erroneous over-unfold


kmuehlbauer:
The proposed changes unfold from first index where kdp < -20 until the end of the beam. If phi is noisy there is a good chance that there will be an over-unfold (not sure how to say), where phi will be greater than by around 360. To overcome this the implementation reduces these pixels by 360. For this the `winlen` is used as search area.

I'm thinking that fixing this should go into a dedicated function. So I'll make some more changes to this PR.

Issue #457, opened 2020-09-20
CI: Use micromamba based installation


kmuehlbauer:
Since `micromamba` is finally available on windows, all CI installations should use `micromamba` to bootstrap the CI conda-forge environment. This will speed up the env creation process and safe CI time.

- [x] Travis CI #464 
- [x] Azure - Linux (MacOSX) #465 
- [x] Azure - Windows - #470
- [x] Appveyor - #470

PR #458, opened 2020-09-21
MNT: use micromamba on travic ci

kmuehlbauer:
closes: #457


kmuehlbauer:
#458 takes over from here

PR #471, opened 2020-11-25
Read truncated radolan

kmuehlbauer:


kmuehlbauer:
@franzmueller I've restructured a bit, but this should work for all products (beside PC, PG where this won't work due to runlength encoding)

PR #481, opened 2021-03-01
FIX: use earthdata login to download srtm data

kmuehlbauer:
- Users now need a NASA Earthdata Login and Password to download SRTM data.
  These credentials have to be specified via environment variables `WRADLIB_EARTHDATA_USER` and `WRADLIB_EARTHDATA_PASS`. 
- Only version 3 data is available from the url https://e4ftl01.cr.usgs.gov/MEASURES/
- Data will automatically be downloaded if not available
- Data will be downloaded to subfolder `geo` of folder where`WRADLIB_DATA` environment variable is pointing to.


kmuehlbauer:
The issue here is, that we can't use our secrets in a Pull Request for security reasons. That means, that the SRTM download can after merge. So, I'll disable the SRTM tests for PullRequests. This is unfortune, but we have no other means for now. There are ways of specially marking a PR to access the secrets but have do be implemented later.

Issue #484, opened 2021-03-03
Cannot find relevant SRTM files for site_coordinates in NW region


gewitterblitz:
Hi,

I was going through the tutorial notebook at this [link](https://github.com/wradlib/wradlib-notebooks/blob/8e3ff5df06479ef7a372ecaabc6b34a6c0d5b079/notebooks/visualisation/wradlib_plot_scan_strategy.ipynb) when I spotted a bug in the `wradlib.io.get_srtm` function. I wanted to use the `plot_scan_strategy` function with terrain using site coordinates for a radar located in Alabama, US. However, wradlib did not like it. Sample code shown below:

```
import wradlib as wrl
import matplotlib.pyplot as pl
import matplotlib as mpl
import os,glob
import warnings
warnings.filterwarnings('ignore')
try:
    get_ipython().magic("matplotlib inline")
except:
    pl.ion()
import numpy as np


nrays = 360
nbins = 150
range_res = 1000.
ranges = np.arange(nbins) * range_res
elevs = [ 0.69763184,  1.3018799 ,  2.1972656 ,  3.2025146 ,  4.2022705 ,
         5.2020264 ,  6.201782  ,  7.201538  ,  8.201294  ,  9.20105   ,
        10.997314  , 12.502441  , 14.002075  , 16.001587  , 18.500977  ,
        20.00061   ]
sitecoords = (-87.34821897, 34.65507505, 178.0) 
beamwidth = 1.0

os.environ["WRADLIB_DATA"] = "/any/folder/of/choice"
os.environ["WRADLIB_EARTHDATA_USER"] = "earthdatausrname"
os.environ["WRADLIB_EARTHDATA_PASS"] = "earthdatapass"

has_data = os.environ.get("WRADLIB_EARTHDATA_USER", False)

if has_data:
    ax = wrl.vis.plot_scan_strategy(ranges, elevs, sitecoords, units='km', terrain=True, az=165)
```
#### Expected Output
Matplotlib image containing radar scan strategy plot with terrain height plotted

#### Actual Output

```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-60-6ea448ce3ae3> in <module>
      1 if has_data:
----> 2     ax = wrl.vis.plot_scan_strategy(ranges, elevs, sitecoords, units='km', terrain=True, az=165)

[...]

PR #485, opened 2021-03-03
replace np.floor in io.get_srtm function

gewitterblitz:
This is in reference to issue #484. Here's the modified wradlib/io/dem.py script to allow terrain computation and SRTM file downloads for site coordinates in NW region.


gewitterblitz:
I added a comment to my OP in issue #484 which means that there is no modification needed in `wradlib.io.dem.get_srtm` function.

PR #500, opened 2021-08-05
FIX: only take valid times when output in `to_odim`

kmuehlbauer:
the wradlib.io.to_odim function in the xarray.py file crashed because the RadarVolume present contained undefined times for certain azimuths. If only the valid_times are considered, the bug is fixed.

@TiemoMathijssen
I've recreated the PullRequest to point to https://github.com/wradlib/wradlib main-branch instead of your forks main branch. The easiest way to PR a branch is to go to the upstream repo (eg. https://github.com/wradlib/wradlib/pulls) and then use "compare across forks" to select your forks feature branch.

### DIFF
--- a/wradlib/io/xarray.py
+++ b/wradlib/io/xarray.py
@@ -950,7 +950,7 @@ def to_odim(volume, filename, timestep=0):
 
         # how group, p. 14 ff.
         h5_ds_how = h5_dataset.create_group("how")
-        tout = [tx.astype("O") / 1e9 for tx in ds.sortby("azimuth").time.values]
+        tout = [tx.astype("O") / 1e9 for tx in ds.sortby("azimuth").time.values[valid_times]]
         tout_sorted = sorted(tout)
 
         # handle non-uniform times (eg. only second-resolution)


kmuehlbauer:
@TiemoMathijssen I've a couple of questions, as there might be more involved. 

How did you create the RadarVolume? 
Did you run any processing on that RadarVolume?

The occurrence of undefined timesteps (`NaT`) can also be an indicator of a problem with reading/processing before. I want to make sure, what actually lead to the `NaT` in your case. Could you wrap up a simple example?

TiemoMathijssen:
The RadarVolume was created in the following way:
The ODIM hdf converter supplied by Rainbow is used to convert the Rainbow volume data to ODIM hdf files. The ODIM hdf file is then loaded into wradlib using the following command:
radvol = wrl.io.open_odim_dataset(filename)

exporting it back to ODIM HDF using the following command fails:
radvol.to_odim('file.h5')

I attached an example ODIM hdf file in a zip archive. [Odim_file.zip](https://github.com/wradlib/wradlib/files/6946465/Odim_file.zip)

kmuehlbauer:
@TiemoMathijssen Thanks for providing the example volume file. So as it turns out, there is a problem with the `_reindex_angle`-function.
https://github.com/wradlib/wradlib/blob/aa4b4cdded2cb0016d012bdaab6abe73f1cd159f/wradlib/io/xarray.py#L1445-L1506

This function makes many assumptions to fix files with eg. missing rays/sectors, double rays etc. It looks like the wradlib default values do not work well with the 90deg sweep of your volume. All other sweeps look fine so far. I'll investigate which parts of this function are the problematic ones.

You can try the following to __not__ run this function by default:

```python
radvol = wrl.io.open_odim_dataset(filename, decode_coords=False)
```

Maybe it's best to make `_reindex_angle` optional and let the user decide if this should be applied according to his data file specifics. Would this be more appropriate than to make assumptions which don't hold?

kmuehlbauer:
So the ultimate source of the problem is the calculation of the angle resolution at this line:

https://github.com/wradlib/wradlib/blob/aa4b4cdded2cb0016d012bdaab6abe73f1cd159f/wradlib/io/xarray.py#L1035

Apparently "median" in connection with the 2-decimals rounding doesn't work for this scan. It leads to and angle resolution of 1.01 which in turn is a wrong assumption for the `_reindex_angle` function. I'll have to give more thought to this.

kmuehlbauer:
@TiemoMathijssen Please note that with the current default behaviour your data will not be read correctly due to the described processing.

kmuehlbauer:
Thanks again @TiemoMathijssen for taking the lead here. Although this fix makes the error go away, it does not solve the underlying problem, unfortunately. #501 fixes the main problem (incorrect calculation of angle resolution) and the data can be read and written correctly. There will be a release in near time. I'll close this in favour of #501.

Issue #526, opened 2021-10-05
Order angles before plotting


kmuehlbauer:
In the function https://github.com/wradlib/wradlib/blob/e5517996c1ba657339e00978e7fb66fd13c94ac2/wradlib/vis.py#L74-L79

the cyclic angle is fixed by adding the first ray at the end again.

In this function `dim0` should be sorted before applying of the fix.

kmuehlbauer:
Yes, the object should be sorted over dimension 'dim0'.

rexdivakar:
i would like to do it please assign it to me

PR #528, opened 2021-10-06
Order angles before plotting #526

rexdivakar:


wradlib/vis.py
```diff
@@ -73,7 +73,7 @@ def __repr__(self):
 
     def fix_cyclic(self):
         rays = self._obj.azimuth
-        dim0 = rays.dims[0]
+        dim0 = rays.dims[0].sort(reverse = False)
```
kmuehlbauer, line 76: @rexdivakar Thanks for taking a stab at this. This is not quite what it should do. I'll try to explain a bit more:

- `self._obj` needs to be sorted over it's `dim0` before `rays` is extracted
- for xarray Dataset you would need to use `.sortby()` method

kmuehlbauer:
@rexdivakar Are you still working on this one?

PR #531, opened 2021-11-04
ENH: Optionalize dependencies

### LINKED ISSUE #506: Make dependencies optional
It would be a nice enhancement, if some/all of the larger wradlib dependencies would be optional.

Every once in a while we get questions concerning wradlib installation and it's dependencies. While the whole installation process has improved tremendously due to the developments of conda-forge and the availability of precompiled wheels, users wonder if dependencies like `GDAL`, `netCDF4`, `h5py`, `h5netcdf`, `matplotlib` and others can be made optional. 

If you consider working with DWD RADOLAN data only `numpy` and `scipy` are needed for reading and processing data. This holds true for any other pure Python-based reader. As wradlib is slowly transitioning to use `xarray` for processing, `xarray` will also be a first class citizen of wradlib and installed by default.

Proposal:

Make dependencies available via `setuptools`'s `extras_require` section.

Currently we can install wradlib's development dependencies via:

```bash
pip install wradlib[dev]
```

There should be these keys:

- `vis` - (`matplotlib`, `cartopy`)
- `io` - (`netCDF4`, `h5py`, `h5netcdf`, `xmltodict`, `requests`)
- `gdal` - (`GDAL`)
- `mp` - (`dask`)
- `all` - (all of the above)
- `dev` - all of the above  plus devel dependencies

Considering the RADOLAN example above we could use:

```bash
pip install wradlib[vis]
```

to get a wradlib installation which can read, process and plot RADOLAN data.

kmuehlbauer:
Resolves #506 

This makes the following dependencies optional, at least when using them in the code:

`dask gdal h5netcdf h5py netCDF4 requests xmltodict cartopy`

If you use pip to install wradlib all dependencies (requirements.txt + requirements_optional.txt) will be installed for the time being. This will change with version 2.0 sometime next year.

If you want to take advantage of this now, you would need to use `--no-deps` when invoking `pip` and install the necessary dependencies beforehand.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,7 +88,7 @@ jobs:
       run: |
         echo $WRADLIB_DATA
         export WRADLIB_DATA=`realpath $WRADLIB_DATA`
-        pytest -n auto --dist loadfile --verbose --doctest-modules --durations=15 --cov-report xml:coverage.xml --cov=wradlib --pyargs wradlib
+        pytest -n auto --dist loadfile --verbose --doctest-modules --doctest-plus --durations=15 --cov-report xml:coverage.xml --cov=wradlib --pyargs wradlib
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v1.2.1
       with:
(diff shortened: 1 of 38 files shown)


kmuehlbauer:
Need to wait for xarray 0.20.1. 0.20.0 is broken ...

PR #536, opened 2021-11-22
MNT: use Python 3.10 conda-forge everywhere, use cartopy

kmuehlbauer:
conda-forge migration for Python 3.10 for the wradlib package-stack is finished. Trying to move CI to Python 3.10 for all OS.


kmuehlbauer:
It looked like everything worked out fine, beside the notebook runs, which had one strange error. Will keep an eye on this.

Issue #591, opened 2022-10-14
Add RADVOR product types RV and RQ to radolan


heistermann:
Long time no see...

The product types RV and RQ are from the RADVOR nowcasting chain:

- RV: nowcast in temporal resolution of 5 minutes ( [example](https://opendata.dwd.de/weather/radar/composite/rv/) )
- RQ: cumulative hourly precipitation depth nowcast for 60 and 120 minutes lead time ( [example](https://opendata.dwd.de/weather/radar/radvor/rq/) )

I guess in order to make `wradlib.io.open_radolan_dataset` work with these data, they need to be added to `wradlib.io.radolan._get_radolan_product_attributes`. Pull request coming up...

Best,
Maik

PR #592, opened 2022-10-14
Add RADVOR product types RV and RQ to radolan.

heistermann:
Please see issue #591 .

### DIFF
--- a/wradlib/io/radolan.py
+++ b/wradlib/io/radolan.py
@@ -834,7 +834,7 @@ def _get_radolan_product_attributes(attrs):
         "EZ",
         "YW",
         "RV",
-        "RQ"
+        "RQ",
     ]:
         pattrs.update(radolan["RR"])
         scale_factor = np.float32(precision * 3600 / interval)
--- a/wradlib/io/radolan.py
+++ b/wradlib/io/radolan.py
@@ -833,6 +833,8 @@ def _get_radolan_product_attributes(attrs):
         "EY",
         "EZ",
         "YW",
+        "RV",
+        "RQ"
     ]:
         pattrs.update(radolan["RR"])
         scale_factor = np.float32(precision * 3600 / interval)


kmuehlbauer:
@heistermann Great to have you back ;-) We need to be careful. With respect to the precision. Are those two really hourly data?

heistermann:
No, RV is 5 minute data.

kmuehlbauer:
Can you link official documents with description? If this is some new format, we might need some adaptions.

heistermann:
And the same for RQ:

`{'producttype': 'RQ',
 'datetime': datetime.datetime(2022, 10, 14, 11, 0),
 'radarid': '10000',
 'datasize': 1620000,
 'formatversion': 5,
 'maxrange': '100 km',
 'radolanversion': '2.29.1',
 'precision': 0.1,
 'intervalseconds': 3600,
 'nrow': 900,
 'ncol': 900,
 'radarlocations': ['asb',
  'boo',
  'drs',
  'eis',
  'ess',
  'fbg',
  'fld',
  'hnr',
  'isn',
  'mem',
  'neu',
  'nhb',
  'oft',
  'pro',
  'ros',
  'tur',
  'umd'],
 'predictiontime': 0,
 'moduleflag': 8,
 'quantification': 1,
 'nodataflag': -9999,
 'secondary': array([], dtype=int64),
 'nodatamask': array([     0,      1,      2, ..., 809997, 809998, 809999]),
 'cluttermask': array([], dtype=int64)}`

heistermann:
Ah, and RV and RQ are also contained in the more verbose [RADOLAN/RADVOR composite format doc](https://www.dwd.de/DE/leistungen/radolan/radolan_info/radolan_radvor_op_komposit_format_pdf.pdf;jsessionid=AEACAAF9B8529EEF4C844D264194761F.live31084?__blob=publicationFile&v=16).

kmuehlbauer:
OK, thanks, I think the RQ can stay in that list, the RV would need to be moved to accompany "WN" above. But I'm not really sure, we need to test that the DataArray has the correct gain/offset.

heistermann:
Why put `RV` together with `WN`? It is not in units of dBZ, but mm of precipitation in a 5 minutes interval, as far as I understand it...similar to `RY`, but analysis plus 24 forecasts from +5 to +120 mins.

kmuehlbauer:
Then we would need to have another "elif" catching RR with 2byte data. RY is only one-byte.

Thegain/offset stuff is needed, because the data is not read until requested from the DataArray. Otherwise we would loose lazy loading.

heistermann:
Isn't RY also two-byte? (`attrs["datasize"]==1620000`)

kmuehlbauer:
Yes, mixed that up with RX :roll_eyes:.

kmuehlbauer:
So, that means everything should be OK as is here in that PR, right? What a fuzz...

heistermann:
Too many product in the RADOLANverse.

heistermann:
But I am confused after looking at the code again. `RY` is considered to be a rainfall rate in mm/h, but it is a depth in mm/5min (see [here](https://www.dwd.de/DE/leistungen/radolan/produktuebersicht/radolan_produktuebersicht_pdf.pdf?__blob=publicationFile&v=9)).

heistermann:
Same holds, in my view, for `RZ`, `EY`, `EZ` and `YW`.

kmuehlbauer:
according to the linked doc RY unit is (1/100mm)/5 min which is a rainrate, right? As in CF-Metadata the Rainrate is in mm/h the output is like this. I'm not sure, the DWD docs are quite irritating with respect to that...

heistermann:
When I read, e.g., `RY` via `read_radolan_composite`, the values are 12 times lower as compared to rading the same dataset via `xr.open_dataset` - which makes sense if we consider the unit RY to be `mm/5min` which should be converted to `mm/h` on the fly. However, that might cause obvious issues if one considers the unit of RY to be `mm` (rainfall_amount/RA )rather than `mm/5min` (rainfall_rate / RR).

kmuehlbauer:
Rainsum/Rainamount would be just mm. This was my thought at least.

heistermann:
Yes, I see.

heistermann:
Maybe this is just a problem on my side...since I, so far, interpreted this as a depth, not as a rate. I was just wondering if users might be confused if they get different number when using read_radolan_composite and xr.open_dataset.

heistermann:
But taking the RADOLAN product docs verbatim, you are right that it is a rainfall_rate.

kmuehlbauer:
Yeah, it turns out that you have to know what you do. Using the DataArray there is clearly described what the unit is, so the confusion should be minimal there. But I have no idea, how to better handle that. I would rely on your expertise here.

heistermann:
You are right! But how do I find the unit of the DataArray if I use `xr.open_dataset`? I cannot find it under "attributes".

heistermann:
Or, to be more specific:
`ry = xr.open_dataset("rydata", engine="radolan")`
`ry.attrs`

gets me

`{'radarid': '10000',
 'formatversion': 3,
 'radolanversion': '2.29.1',
 'radarlocations': ['asb',
  'boo',
  'ros',
  'hnr',
  'umd',
  'pro',
  'ess',
  'fld',
  'drs',
  'neu',
  'nhb',
  'oft',
  'eis',
  'tur',
  'isn',
  'fbg',
  'mem']}`

So is it possible that the main portion of metadata is lost on the way? Or maybe I should read the docs first...

kmuehlbauer:
Yes, not needed metadata or due to the fixing unreasonable metadata is removed. 

But the units are here in the DataArray not the Dataset: `ry.RY.attrs`.

kmuehlbauer:
commits squashed and added to #594

PR #597, opened 2022-10-24
Changes in # read metadata under get_radiosonde()

JulianGiles:
A bug in the original code in misc.py throws an error when the station being loaded with get_radiosonde() does not have the 'Station identifier' parameter. The problem arises because an iterator of parameters is created and then some transformations are made considering the indexing of the elements of the iterator, under the # read metadata section. When the 'Station identifier' parameter (which is the first element) is missing, the wrong parameter is selected and the error occurs. In this commit, changes in the # read metadata section allow for transforming the set of parameters of metadata referring to the keywords instead of the indexes, in order to avoid the problem of not having the 'Station identifier' parameter.

### DIFF
--- a/wradlib/io/misc.py
+++ b/wradlib/io/misc.py
@@ -194,11 +194,11 @@ def get_radiosonde(wmoid, date, cols=None):
         k, v = row.split(":")
         k = k.strip()
         v = v.strip()
-        if i == 2:
+        if k == 'Station number':
             v = int(v)
-        elif i == 3:
+        elif k == 'Observation time':
             v = dt.datetime.strptime(v, "%y%m%d/%H%M")
-        elif i > 3:
+        elif i > 2:
             v = float(v)
         meta[k] = v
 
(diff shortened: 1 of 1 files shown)


wradlib/io/misc.py
```diff
@@ -194,11 +194,11 @@ def get_radiosonde(wmoid, date, cols=None):
         k, v = row.split(":")
         k = k.strip()
         v = v.strip()
-        if i == 2:
+        if k == 'Station number':
```
kmuehlbauer, line 197: Please use double ticks " here and below. Otherwise the linter doesn't like it.

kmuehlbauer (changes requested):
This is looking good to me. Making the linter happy and this can go in.

kmuehlbauer (approved):

kmuehlbauer:
I'm wondering why the codecov-bot doesn't announce the coverage statistics.

Thanks @JulianGiles for the fast PR and welcome as a first time contributor! Hope to see you more around.

Issue #602, opened 2022-11-08
Add specified percentage parameter to clutter.histo_cut


overeem11:
This identification algorithm uses the histogram of temporal accumulated rainfall. It iteratively detects classes whose frequency falls below a specified percentage (1% by default) of the frequency of the class with the biggest frequency and remove the values from the dataset until the changes from iteration to iteration falls below a threshold. This specified percentage is currently hard-coded, which limits application of the algorithm.

#### MCVE Code Sample

```python
for i in range(index, -1, -1):
    if n[i] < (n[index] * 0.01):
          break

 for j in range(index, len(n)):
     if n[j] < (n[index] * 0.05):
           break
```

#### Expected Output
This cannot be reproduced with a few lines of code, but the output will remain the same if 0.01 would be supplied as a parameter (specified percentage) to the histo_cut function. Output will of course change when another value would be chosen.

#### Problem Description

Recommendation is to change the hard-coded default value of 0.01 (1%) in the histo_cut function in a parameter, so to add a parameter to wradlib.clutter.histo_cut. For OPERA radar data a value of 0.05 seems to work quite well, and this is currently only possible by changing the hard-coded value of 0.01 in 0.05 in the Python script "clutter.py" and by loading this script instead of loading the wradlib package. The 0.01 in the example above should then be replaced by a parameter, e.g. threshold, which should then be supplied as argument to the histo_cut function. The parameter would then occur in the two for loopes provided above.

#### Version

<details><summary>Output of <tt>wrl.show_versions()</tt></summary>

</details>

kmuehlbauer:
Thanks for opening this feature request @overeem11. That seems totally reasonable to make that value available as parameter.

One question though, would it make sense to provide two parameter, one for each loop (upper/lower bound)?

I think the code could also need some refreshing. I hope to free some cycles to get this done soon.

overeem11:
Thanks. Good idea to make a two parameter solution.

Van: Kai Mühlbauer ***@***.***>
Verzonden: woensdag 9 november 2022 07:42
Aan: wradlib/wradlib ***@***.***>
CC: Overeem, Aart (KNMI) ***@***.***>; Mention ***@***.***>
Onderwerp: Re: [wradlib/wradlib] Add specified percentage parameter to clutter.histo_cut (Issue #602)


Thanks for opening this feature request @overeem11<https://github.com/overeem11>. That seems totally reasonable to make that value available as parameter.

[...]

PR #603, opened 2022-11-09
Histo cut and xarray backends

kmuehlbauer:
This PR adds parametrization to wradlib.clutter.histo_cut and adds description to wradlib xarray backends.

Fixes: #602


kmuehlbauer:
Should be rebase-merged!

PR #639, opened 2023-09-11
Fix wrong prediction_time variable in radolan xarray.Dataset

Hoffmann77:
Changed the dt.timedelta argument to the 'minutes' for a correct calculation of 'prediction_time'.

radvor

### DIFF
--- a/wradlib/io/radolan.py
+++ b/wradlib/io/radolan.py
@@ -1150,7 +1150,9 @@ class _radolan_file:
 
         pred_time = attrs.get("predictiontime", None)
         if pred_time is not None:
-            pred_time = np.array([(raw_time + dt.timedelta(pred_time)).timestamp()])
+            pred_time = np.array(
+                [(raw_time + dt.timedelta(minutes=pred_time)).timestamp()]
+            )
             pred_time_var = WradlibVariable(
                 "prediction_time", data=pred_time, attrs=time_attrs
             )


kmuehlbauer:
@Hoffmann77 Thanks for the PR. Could you link a document as reference here? 

Otherwise looking good 👍.

wradlib/io/radolan.py
```diff
@@ -1150,7 +1150,9 @@ def _read(self):
 
         pred_time = attrs.get("predictiontime", None)
         if pred_time is not None:
-            pred_time = np.array([(raw_time + dt.timedelta(pred_time)).timestamp()])
+            pred_time = np.array(
+                [(raw_time + dt.timedelta(minutes=pred_time)).timestamp()]
```
kmuehlbauer, line 1154:
```suggestion
                [raw_time + dt.timedelta(minutes=pred_time)], dtype=np.datetime64
```

kmuehlbauer (changes requested):
@Hoffmann77 I've checked with the compiled docs at https://wradlib--639.org.readthedocs.build/en/639/notebooks/fileio/radolan/radolan_showcase.html#RADOLAN-RADVOR. Although your changes are very valid, we would need to directly create numpy datetime64 instead going over the `timestamp()`. 

I've made a suggestion to fix this, then we can merge. Please also add an entry to doc/release_notes.md to make this fix visible.

kmuehlbauer:
Thanks @Hoffmann77 for catching and fixing this! Welcome to the world of wradlib. The docs will be rendered correctly soon.

PR #670, opened 2024-07-04
MNT: update CI, numpy2

kmuehlbauer:


wradlib/georef/vector.py
```diff
@@ -479,6 +479,6 @@ def get_centroid(polyg):
         x and y coordinate of the centroid
 
     """
-    if not type(polyg) == ogr.Geometry:
+    if type(polyg) is not ogr.Geometry:
```
mats-knmi, line 482: These `is not` checks should be replaced with isinstance checks as well I think. Or do we want to be absolutely sure that `trg` is of type `Geometry` and not a subtype of `Geometry`?
```suggestion
    if not isinstance(polyg, ogr.Geometry):
```

wradlib/zonalstats.py
```diff
@@ -340,7 +342,7 @@ def _get_intersection(self, *, trg=None, idx=None, buf=0.0):
             raise TypeError("Either `trg` or `idx` kwargs must be given!")
 
         # check for geometry
-        if not type(trg) == ogr.Geometry:
+        if type(trg) is not ogr.Geometry:
```
mats-knmi, line 345: Same here
```suggestion
        if not isinstance(trg, ogr.Geometry):
```

wradlib/georef/vector.py
kmuehlbauer: Obviously missed those. Let's try it.

kmuehlbauer:
@mats-knmi Thanks for engaging here. I'm going to merge this right after getting green lights. 

The other deprecation warnings can be handled in subsequent PR's.

The notebook-test is non-trivial to solve as `wetterdienst`-package latest version has diverged quite a bit from the version I've pinned (which only works on numpy 1, sigh). They are also pinning dependencies quite strict which prevents current usage. So need to think about that a bit.

kmuehlbauer:
Most likely 0.49 wasn't pinned for numpy?

I do not like the strict pinning scheme over those many dependencies. In connection with conda- forge you end up in unresolvable environments pretty fast.

mats-knmi:
Yeah strict dependencies just doesn't work with python since only one version of a package may be installed at once.

As far as I can see wetterdienst `0.49.0` also has a dependency on numpy < 2, so I don't understand how the github actions conda environment creation step can install these 2 dependencies side by side.

kmuehlbauer:
I have no idea, the magic of GitHub. : grimacing:

PR #671, opened 2024-08-01
numpy 2.0 compatibility

mats-knmi:
I noticed that some parts of wradlib currently do not work with numpy 2.0. I fixed these issues in the code and tried to make the unit tests github actions succeed after that. This unearthed some other issues related to the newest gdal version, so I had to pin the gdal version for the test environments to 3.8.

Now all tests seem to succeed, except for 2 tests in the georeferencing module. The outcome differs more than 3 decimals from the expected output. This probably also has something to do with the new gdal version or maybe the new numpy version, since I changed nothing functionally in the code.

### DIFF
--- a/ci/requirements/notebooktests.yml
+++ b/ci/requirements/notebooktests.yml
@@ -10,7 +10,7 @@ dependencies:
   - coverage
   - dask
   - deprecation
-  - gdal
+  - gdal=3.8
   - geopandas
   - h5py
   - h5netcdf
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -94,6 +94,7 @@ exclude = [
     "docs",
 ]
 
+[tool.ruff.lint]
 # E402: module level import not at top of file
 # E501: line too long - let black worry about that
 # E731: do not assign a lambda expression, use a def
@@ -114,9 +115,10 @@ select = [
     "I",
     # Pyupgrade
     "UP",
+    "NPY201",
 ]
 
-[tool.ruff.isort]
+[tool.ruff.lint.isort]
 known-first-party = ["wradlib"]
 
 [tool.pytest.ini_options]
--- a/wradlib/georef/xarray.py
+++ b/wradlib/georef/xarray.py
@@ -109,18 +109,18 @@ def create_xarray_dataarray(
 
     if phi is None:
         if sweep_mode == "azimuth_surveillance":
-            phi = np.arange(data.shape[0], dtype=np.float_)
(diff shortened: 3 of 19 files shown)


mats-knmi:
Oh oops I see now that this is actually a duplicate of https://github.com/wradlib/wradlib/pull/670. I will close this pull request.

kmuehlbauer:
@mats-knmi No worries, I'm back from vacation this week and didn't had the time yet to pick this up. As you have already put quite some work in this maybe we can combine bot PR and get it over the finish line. WDYT? 

I'll have some questions inline.

pyproject.toml
```diff
@@ -114,9 +115,10 @@ select = [
     "I",
     # Pyupgrade
     "UP",
+    "NPY201",
```
kmuehlbauer, line 118: Does this enable ruff to explicitely cover for numpy2 deprecations?

wradlib/georef/xarray.py
```diff
@@ -109,18 +109,18 @@ def create_xarray_dataarray(
 
     if phi is None:
         if sweep_mode == "azimuth_surveillance":
-            phi = np.arange(data.shape[0], dtype=np.float_)
+            phi = np.arange(data.shape[0], dtype=np.float64)
```
kmuehlbauer, line 112: I'm not sure but would `float` be enough? Or is it better to explicitely ask for `float64`. I'm not that deep into this.

ci/requirements/notebooktests.yml
```diff
@@ -10,7 +10,7 @@ dependencies:
   - coverage
   - dask
   - deprecation
-  - gdal
+  - gdal=3.8
```
kmuehlbauer, line 13: I think I've addressed the gdal issues over at my PR. You might try it out.

pyproject.toml
mats-knmi: Yes indeed this is what I used to quickly fix all the numpy2 issues.

This actually enabled me to simply run `ruff check --fix` and that fixed all numpy 2.0 issues automatically. This rule could actually be removed now, since it is mostly just useful during the migration.

ci/requirements/notebooktests.yml
kmuehlbauer: The branch is here now: https://github.com/kmuehlbauer/wradlib/tree/update-ci-safe

mats-knmi: yeah indeed I saw that, better than just pinning to an old version

PR #684, opened 2024-12-10
FIX manual garbage collector is needed in test with object creation

egouden:
Running the tests was not possible on a modest machine with 4GB of RAM running debian 12. It appeared quickly that the memory was growing after each separate test.

After several hours of investigation it seems that it is related to the [caching of fixtures](https://github.com/pytest-dev/pytest/discussions/10387).

This quick fix allows reducing the memory footprint by cleaning manually the most demanding part of the tests.

### DIFF
--- a/docs/release_notes.md
+++ b/docs/release_notes.md
@@ -7,6 +7,10 @@ Please note that {{wradlib}} releases follow [semantic versioning](https://semve
 
 You can install the latest {{wradlib}} release from PyPI via ``$ python -m pip install wradlib`` or specific version via ``$ pip install wradlib==x.y.z``. The recommended installation process is described in {doc}`installation`.
 
+## Development version
+
+* FIX: manual garbage collector is needed before each test
+
 ## Version 2.3
 
 This is a pure maintenance release, following the merge of datatree into xarray codebase. This release will pin xradar >= 0.8.0 and xarray >= 2024.10.0 and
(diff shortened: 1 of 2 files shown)


wradlib/tests/test_georef.py
```diff
@@ -807,6 +808,7 @@ class Data:
             ]
         )
 
+    gc.collect()
```
kmuehlbauer, line 811: Can you please add a comment here, so that we keep track of why we garbage collect before yield.

kmuehlbauer (changes requested):

wradlib/tests/test_georef.py
egouden: A comment has been added.

kmuehlbauer:
@egouden The test runs do not indicate increased run-time. So I think this is good to go. Please add an entry to https://github.com/wradlib/wradlib/blob/main/docs/release_notes.md under a new section "Development Version". Thanks!

docs/release_notes.md
```diff
@@ -7,6 +7,10 @@ Please note that {{wradlib}} releases follow [semantic versioning](https://semve
 
 You can install the latest {{wradlib}} release from PyPI via ``$ python -m pip install wradlib`` or specific version via ``$ pip install wradlib==x.y.z``. The recommended installation process is described in {doc}`installation`.
 
+## Development version
+
+* FIX: manual garbage collector is needed before each test
```
kmuehlbauer, line 12:
```suggestion
* FIX: manual garbage collector is needed before each test ({pull}`684`) by {at}`egouden`
```

kmuehlbauer (approved):

kmuehlbauer:
Thanks @egouden!

PR #685, opened 2024-12-11
ENH use only one data variable for testing

egouden:
A dataset containing many variables is used in test_cross_session_ppi. This needs more than 1GB of memory.

Using only DBZH significantly reduces the memory footprint of the test.

### DIFF
--- a/wradlib/tests/test_util.py
+++ b/wradlib/tests/test_util.py
@@ -333,14 +333,16 @@ def test_find_bbox_indices(bb_data):
 @requires_data
 def test_cross_section_ppi():
     file = util.get_wradlib_data_file("hdf5/71_20181220_061228.pvol.h5")
+    # only load DBZH for testing
+    keep_vars = ["DBZH", "sweep_mode", "sweep_number", "prt_mode", "follow_mode"]
     # load all sweeps manually and merge them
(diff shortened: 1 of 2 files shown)


wradlib/tests/test_util.py
```diff
@@ -334,13 +334,14 @@ def test_find_bbox_indices(bb_data):
 def test_cross_section_ppi():
     file = util.get_wradlib_data_file("hdf5/71_20181220_061228.pvol.h5")
     # load all sweeps manually and merge them
+    keep_vars = ["DBZH", "sweep_mode", "sweep_number", "prt_mode", "follow_mode"]
```
kmuehlbauer, line 337:
```suggestion
    # only load DBZH for testing
    keep_vars = ["DBZH", "sweep_mode", "sweep_number", "prt_mode", "follow_mode"]
    # load all sweeps manually and merge them
```

kmuehlbauer (changes requested):
Thanks @egouden, it's reasonable to reduce the test to only one moment. I've made a small suggestion to reorder comments. Please also add an entry to docs/release_notes.md.

kmuehlbauer (approved):
Thanks @egouden!

PR #691, opened 2025-02-06
ADD: transform sweep into a composite raster image

egouden:
This contribution has two goals:
- to improve the accuracy of spatial transformations in compositing
- to provide a higher level interface for compositing based on the sweep dataset object

Unit tests are provided to verify that the code can be run without errors. [A few examples are provided](https://github.com/egouden/wradlib-notebooks/blob/new_composite_examples/notebooks/compositing.ipynb) to actually verify the quality of the new feature.

Some notes:
- this is a direct application of the work on improved interpolation methods performed several years ago
- the integration with existing gridding functions and Xarray accessors could be considered in a more general discussion at a later stage
- georeferenced composite files can now be produced easily


kmuehlbauer:
@egouden This looks like something most wanted for years now :-) I'll have a closer look the next days.

> * Does the small convenience of using rioxarray justifies adding it as dependency?

Please use `xradar.georeference.add_crs` instead (maybe via pipe for now). There is also an accessor at xradar, but it doesn't allow kwargs yet.

egouden:
This is back to "draft mode" as the georeferencing is not recognised by GDAL.

egouden:
> Thanks for the quick reply. This is back to "draft mode" since the georeferencing is actually not recognised by GDAL.

I spent hours to find how to make GDAL happy. But now I guess it is ready for review.

kmuehlbauer (changes requested):
@egouden Thanks for taking this on and for your patience, I've finally come to this now. 

This is something which has been asked every once in a while now. The current available implementation for creating cartesian grids  (`comp.togrid`) isn't very user friendly. I'm open for adding this enhancement here. I'm a bit concerned about the usage of kwargs (I know wradlib has them all over the place) and I think we should use them sparsely and with care.

egouden:
Thank you for the suggestions !

I increased the modularity and clarity. Some reasonable default behaviors have been chosen.

I think the main use case for RectBin is the transformation/interpolation of irregular radar data into regular raster/grid. And filling the gaps between the rays is part of the job. Accordingly since no data might be present after radar quality control, the default function is now np.nanmean. In my opinion, this function should be useful from a radar perspective and not just be a wrapper to a scipy function. This is still an option for backward compatibility but I would make it standard in the next major version.

The notebooks helps to understand the new feature. They are also the actual tests. What do you think about it?

kmuehlbauer:
@egouden Thanks for pushing this further. I've taken the chance to reimplement `sweep_to_raster` using `apply_ufunc` such that it works for Dataset and DataArray inputs. 

We could still improve this, if we would find a neat solution to integrate `raster`/`transform` into one object or to rework the interpolator classes to be more xarray-friendly.

Please let me know, if you are d'accord with this reimplementation. Then I'd move forward here and merge/release.

kmuehlbauer:
Thanks @egouden! Great addition!

PR #692, opened 2025-02-10
FIX proj data package is needed for geoid heights

egouden:
Coordinate transformation from ellipsoid to geoid heights fails if the needed data are missing. This is the case on the current test configuration in windows.

This is fixed (quickly) by installing the proj-data package explicitly in conda.

Note that GDAL does not raise an error by default and simply returns the same coordinates. This is caught by a unit test which is now working for the windows configuration.

### DIFF
--- a/ci/requirements/unittests.yml
+++ b/ci/requirements/unittests.yml
@@ -20,6 +20,7 @@ dependencies:
   - numpy
   - matplotlib-base
   - pip
+  - proj-data
   - pytest
   - pytest-cov
   - pytest-doctestplus
--- a/wradlib/tests/test_georef.py
+++ b/wradlib/tests/test_georef.py
@@ -3,7 +3,6 @@
 # Distributed under the MIT License. See LICENSE.txt for more info.
 
 import gc
-import sys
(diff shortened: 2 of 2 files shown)


kmuehlbauer:
Thanks @egouden for trying to fix this. Although this fixes the CI issues, this is not the proper solution as this will introduce approx. 750MB of additional data downloaded in the CI.

In #693 I'm experimenting with setting PROJ_NETWORK = "ON" to make the grid data available per download. For some reason although passing the test suite windows returns with non zero code failing CI. We might need to add some more environment variables for proper working in any OS/python versions.

kmuehlbauer:
I agree with your reasoning, but I think the proper solution is to make the grid available by setting env variables and checking availability of grids before usage instead of silently returning wrong results.

egouden:
The main cause of this is the breaking change in [proj7]( https://www.osgeo.org/foundation-news/proj-7-0-0/)

egouden:
It seems you have the proper solution indeed. If you are stuck, I can give it a try.

PR #693, opened 2025-02-10
Use proj network, update CI and code

kmuehlbauer:


kmuehlbauer:
I found some more issues where tests are not declared with @requires_data. There may be more fixtures which are ill-assigned or missing. Anyway, it looks like there is an issue with some special functions I will track down later.

kmuehlbauer:
@egouden I'll not have time until next week, full schedule.

kmuehlbauer:
@egouden I've finally got around this. Main points:

- removed the Numpy 1 CI build. Numpy 1 gets out of service on 17 Sep 2025, hope this is OK
- use '--dist loadfile' and nomkl in windows builds (the first one magically solved the segmentation fault)
- add pooch to CI environments (for wradlib-data dependencies)
- yield instance of dataclass in the former failing test
- simplifiy geoid crs creation
- unskip/unfail tests

Looks like loading projection data on the fly is working.

wradlib/tests/test_georef.py
```diff
@@ -37,9 +35,9 @@
 
 
 # this ensures objects from previous tests are cleaned
-@pytest.fixture(autouse=True)
-def ensure_gc():
-    gc.collect()
+# @pytest.fixture(autouse=True)
```
kmuehlbauer, line 38: @egouden before I merge this, can you check, if this garbage collection fixture is still needed?

PR #698, opened 2025-02-28
ENH use wradlib-data package by default

egouden:
The recently introduced wradlib-data package is easier to use than the WRADLIB_DATA environment variable. This change aims to use only the former across the wradlib package. The latter is kept for backward compatibility but it could be removed in the next major version.

Additionally, the wradlib-data package has been added as a mandatory requirements given the benefits. The lack of data is therefore not checked anymore.


egouden:
This is passing the tests locally but fails on the CI. I hope this is just a configuration issue.

wradlib/util.py
```diff
@@ -584,26 +585,25 @@ def has_geos():
...
+        wrl_data_path = DATASETS.abspath
+
+    if not os.path.exists(wrl_data_path):
+        os.makedirs(wrl_data_path)
+
     return wrl_data_path
 
 
 def get_wradlib_data_file(relfile):
     data_file = os.path.abspath(os.path.join(get_wradlib_data_path(), relfile))
```
kmuehlbauer, line 605: This will not work, if the wanted file isn't yet downloaded. We will have to explicitly 

```python
from wradlib_data import DATASETS
data_file = DATASETS.fetch(relfile)
```

to retrieve the file from the GH repo into a path which can be overridden by `WRADLIB_DATA`.

wradlib/util.py
```diff
@@ -584,26 +585,25 @@ def has_geos():
...
-        raise OSError(
-            "`WRADLIB_DATA` environment variable not set. Please set `WRADLIB_DATA` "
-            "pointing to the location of `wradlib_data` repository on the filesystem."
+
+    if "WRADLIB_DATA" in os.environ:
+        wrl_data_path = os.environ["WRADLIB_DATA"]
+        warn(
+            "usage of the WRADLIB_DATA environment variable will be deprecated in the future in favor of the wradlid_data package",
+            FutureWarning,
         )
```
kmuehlbauer, line 594: Usage of `WRADLIB_DATA` will not be deprecated. It will still be used, so that the user can specify the location where `wradlib-data`-package will store the data. 

We should test for availability of `wradlib-data`-package and warn if it is not yet used.

wradlib/util.py
kmuehlbauer: This might be the root cause of the breaking tests.

kmuehlbauer (changes requested):
@egouden I've left some comments and suggestions on how to move this forward. Generally this goes into the right direction, advertising the usage of `wradlib-data`-package instead of downloading the complete repo (as users had to do before).

wradlib/util.py
egouden: Thanks for the review @kmuehlbauer. The comments have been integrated. You can make further changes if needed.

egouden:
Thanks for the review @kmuehlbauer. The comments have been integrated. You can make further changes if needed.

egouden:
I guess this is due to the fact that `requires_data` has been removed. Should data not be mandatory to perform tests?

kmuehlbauer:
So, just for the sake of completeness, here is where `wradlib-data`/`WRADLIB_DATA`-env var is used within wradlib ecosystem:

### wradlib-data

- wradlib tests
- wradlib-notebooks

### WRADLIB_DATA

- [util.get_wradlib_data_path](https://github.com/wradlib/wradlib/blob/cae391d45a26450c389af7329ba5e583a66eb3de/wradlib/util.py#L586C5-L586C26) (used for getting `WRADLIB_DATA`path)
- [util.get_wradlib_data_file](https://github.com/wradlib/wradlib/blob/cae391d45a26450c389af7329ba5e583a66eb3de/wradlib/util.py#L598) (used for extracting file from `WRADLIB_DATA` path)
- in [io.dem/get_srtm](https://github.com/wradlib/wradlib/blob/cae391d45a26450c389af7329ba5e583a66eb3de/wradlib/io/dem.py#L150) we use the above to get a writable (geo)-folder for downloading srtm tiles.

With the change to pooch-based wradlib-data we are able to deprecate and remove both of the above functions. They haven't been part of the public API, anyway. 

To move forward we should warn about that change in these two functions (how to retrieve the files for tests/notebooks). For the SRTM-part we would still rely on a given WRADLIB_DATA env var (which can be the same one from wradlib-data).

`wradlib-data` should not be a dependency of `wradlib`, but only a test/doc-dependency. So I'd suggest to add this to `requirements_devel.txt` to make this more visible.

egouden:
Could we use wradlib-data by default for STRM data as well?

kmuehlbauer:
> Could we use wradlib-data by default for STRM data as well?

If wradlib-data is installed it's path will be used for SRTM data via util._get_wradlib_data_path().

If you have WRADLIB_DATA env var set, this path is directly used (since it would also be retrieved via wradlib-data). The main thing is, that users do not have to install wradlib-data if they want to use SRTM data.

kmuehlbauer (approved):

kmuehlbauer:
@egouden I think, this is good to go now.

PR #703, opened 2025-04-07
ENH: practical guide for contributing

egouden:
This adds a practical guide for contributing including the installation of all the dependencies on various platform.

I have the following issues:

- I think testing the notebooks is useful as well. But I do not know how to do it automatically.

- The use of pip inside conda environement works but it is not clean. Is it possible to define a developement version of wradlib in the conda structure?

- The installation of the latest version of gdal is tricky on Windows. I spend several days on it. Installing the latest precompiled binnaries is easy. But I got DLL issues  when using it together with coverage. I looked at a static build but I am still facing remaining errors.

This is mainly based on my own experience contributing. @kmuehlbauer I am looking forward your input as maintainer.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,36 +1,236 @@
-# Contributing to wradlib
 
-The wradlib project encourages everyone to contribute to the developement of wradlib. This paper describes the general guidelines for contributing to wradlib. There is more information on [wradlib's community effort](https://wradlib.org/community) at the project's main website [wradlib.org](https://wradlib.org).
+# Contributing
 
-In general wradlib uses GitHub Pull Requests for contribution. Second, the [wradlib developers](CONTRIBUTORS.txt) have GitHub Commit Access to the project's main repository.
+This is a practical guide for contributing. Please check the [documentation](docs/dev_guide.md) for more general information.
 
-## Code of Conduct
+## 1a Install environement using conda
+- Install miniconda3 on Windows and start Anaconda Powershell Prompt:
+``` powershell
+winget install --id=Anaconda.Miniconda3 -e --silent
+exit
+```
+- Create development environement (use "wradlib-dev" to keep wradlib for release):
+```powershell
+conda config --add channels conda-forge
+conda create -n wradlib
+conda activate wradlib
+```
+- Install wradlib package:
+```bash
+conda install wradlib
+```
+- Install gdal optional packages:
+```bash
+conda install libgdal-hdf5
+conda install libgdal-netcdf
+```
 
-Contributors to wradlib are expected to act respectfully toward others in accordance with the [OSGeo Code of Conduct](http://www.osgeo.org/code_of_conduct).
+## 1b Install environement using pip on Windows
 
-## Contributions and Licensing
+- Install latest version of python
+``` powershell
+winget install -e --id Python.Python.3.13
+exit
+```
+- Create virtual environement:
+``` powershell
+python -m venv $env:USERPROFILE/.venvs/wradlib
+Set-ExecutionPolicy -Scope CurrentUser -ExecutionPolicy RemoteSigned
+New-Item -Path $PROFILE -ItemType File -Force
+'Set-Alias wradlib "$env:UserProfile\.venvs\wradlib\Scripts\activate.ps1"' | Out-File -Append $PROFILE
+exit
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -29,8 +29,214 @@ used wradlib as well as Python and OS version, error messages and instructions t
...
-To keep the code as clean as much all code contributions should be in line with [PEP 8](https://www.python.org/dev/peps/pep-0008/) conventions.
+For this purpose, you need to have a [GitHub Account](https://github.com/signup/free) and [Git](https://git-scm.com/). Then fork wradlib, make changes and create a pull request.
+
+## Guide
+
+This is a practical step by step guide. The commands should work in your operating system terminal. Please adapt it to your needs.
+
+### 1a Install development environement using conda
+
+- Install [Anaconda or Miniconda](https://www.anaconda.com/download/)
```
kmuehlbauer, line 40: Maybe just refer to the general `Installation`-section for the conda based install.

kmuehlbauer (changes requested):
@egouden Thanks for this nice writeup. 

Unfortunately the doc is still a bit out-of-sync since moving it back to the wradlib-main repo. So we have `CONTRIBUTING.md` (where you have placed these changes) and we have `docs/dev_guide.md` which is rendered in the documentation.

My suggestion would be to reduce `CONTRIBUTING.md` to a paragraph linking to the dev_guide withing the documentation and incorporate your additions there. Thanks!

egouden:
I suggest to keep a simple technical guide in CONTIBUTING.md. A more general guide providing context is useful in the documentation. It needs to be revised.

The installation now works on windows with latest cmake (not sure if this was the actual culprit). The idea is to have the latest software for development. I will try spack as package manager for linux to get the same result.

CONTRIBUTING.md
egouden: I actually propose a minimal working installation of conda. Please check the new specific guide for windows(the linux guide is coming). The installation of dev and opt dependencies remain an open question. A reference to the more elaborated installation guide can be made for better context.

docs/dev_guide.md
```diff
@@ -1,5 +1,37 @@
...
+
+All contributions shall comply to the project [license](LICENSE.txt).
+
+## GitHub
+
+Source code, testing, bug tracking as well as docs and wiki are handled via GitHub.
+
+## Documentation
+
+* wradlib's documentation source is contained in [wradlib-docs](https://github.com/wradlib/wradlib-docs) repository, and mainly written in reStructuredText.
```
kmuehlbauer, line 21: That's not the case anymore. The docs live in the main-repo again since #619.

kmuehlbauer:
> I suggest to keep a simple technical guide in CONTRIBUTING.md. A more general guide providing context is useful in the documentation. It needs to be revised.

I tend to agree, but we should try to avoid duplicating documentation. So when revising, we should import CONTRIBUTING.md verbatim into the wanted destination (via include-directive).

docs/dev_guide.md
kmuehlbauer, line 21:
```suggestion
* wradlib's documentation source is contained in the [docs-folder](https://github.com/wradlib/wradlib/tree/main/docs) of the wradlib repository, and mainly written in reStructuredText.
```

docs/dev_guide.md
```diff
@@ -1,5 +1,37 @@
 # Contributing Guide
 
+The wradlib project encourages everyone to contribute to the developement of wradlib. This paper describes the general guidelines for contributing to wradlib. There is more information on [wradlib's community effort](https://wradlib.org/community) at the project's main website [wradlib.org](https://wradlib.org).
+
+In general wradlib uses GitHub Pull Requests for contribution. Second, the [wradlib developers](CONTRIBUTORS.txt) have GitHub Commit Access to the project's main repository.
```
kmuehlbauer, line 5:
```suggestion
In general wradlib uses GitHub Pull Requests for contribution. Second, the [wradlib developers](https://github.com/wradlib/wradlib/blob/main/CONTRIBUTORS.txt) have GitHub Commit Access to the project's main repository.
```

kmuehlbauer (approved):

kmuehlbauer:
Thanks again, @egouden!

PR #704, opened 2025-05-16
FIX: binned transform filling was not correct

egouden:
This is a fix for several issues related to gap filling in the default transformation of sweep to raster by binning:

- the computation of the radar mask was expecting a boolean matrix from extract_circle, but a list of 1D indices is returned instead
- the center was using geographic coordinates, which is not appropriate for extracting radar circle
- the geographic CRS were not transformed to projected coordinates for proper circle computation

The [notebook](https://github.com/egouden/wradlib-notebooks/blob/3a0773e4950591ade9c16c688d4aa8545a7c78c4/notebooks/compositing/max_reflectivity.ipynb) failed to catch the first issue because a masked dataset with nans was used following a range selection below 200km. This actually provided correct result. The dataset is now properly cut from the full dataset, exposing the issue. A new geographic raster has been added in the new version of the notebook.

### DIFF
--- a/wradlib/comp.py
+++ b/wradlib/comp.py
@@ -30,7 +30,7 @@ from functools import singledispatch
 import numpy as np
 from xarray import DataArray, Dataset, apply_ufunc, broadcast, concat, set_options
 
-from wradlib.georef import wkt_to_osr
+from wradlib.georef import get_radar_projection, reproject, wkt_to_osr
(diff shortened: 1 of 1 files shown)


egouden:
Temporary remark: the CI is failing due to external reasons related to downloading from wradlib-data.

kmuehlbauer (approved):

PR #709, opened 2025-08-08
use tolerance for regular size when creating raster

egouden:
Due to limited precision, providing a regular size can results in a raster array with an extra row and column.

This fix allows for a tolerance on the regular size precision.


egouden:
I followed another approach in #711.

PR #711, opened 2025-08-22
ENH enhance georeferenced raster creation

egouden:
* follow GIS convention for bounds
* enforce precision
* option to snap bounds/resolution
* use arcsecond unit for geographic CRS
* use rioxarray for metadata handling


egouden:
I am pointing temporary to new notebooks due to breaking change (bounds order).

Issue #712, opened 2025-08-25
integrate notebooks


egouden:
I see several reasons to integrate the notebooks in the main wradlib repository:

- integrate notebook in testing
- avoid breaking change
- better visibility for users
- follow xradar strategy

kmuehlbauer:
Thanks Edouard for creating this issue.

I agree. The root causes to have a dedicated notebooks-repo do not exist any more and the handling now is a bit of a pain. We have brought the docs back already in v2.0.0 for the same reason and I think it's time to also include the notebooks.

This might look like an easy task, but we should discuss this in more detail. 

What I would like bring into the discussion:

- use [MyST]](https://mystmd.org/) markdown for writing notebooks instead of regular notebooks
- drop Python intro, instead link to other resources (NumPy, SciPy etc)
- restructure the tutorials (eg. dedicated Getting started section with low barrier simple examples)
- drop backend specific notebooks which are already handled in xradar
- add a gallery reel

kmuehlbauer:
Can you give a bit more detail how you envision the pipeline where the notebook rendering does  not run on ReadTheDocs.

The current pipeline works quite well and if we have the new notebooks in wradlib main, we can still use it. All source files in one place, only the rendered output over at wradlib-notebooks.

I can add a PR with my suggestion later this week

kmuehlbauer:
@egouden I'm on it, but this will take a bit more time as expected. I'm following multiple path's now, and hope to have a PR ready early next week. Thanks for your patience.

egouden:
This looks quite good. A lot of work already. I suggest to close the general issue and open new issues for most wanted features. Thx @kmuehlbauer  !

kmuehlbauer:
@egouden I'm fine with that approach! Closing.

PR #714, opened 2025-09-07
MystNB text-only notebooks

kmuehlbauer:
This introduces MystNB text-only notebooks into wradlib repo.

All notebooks have been more or less copied over from https://github.com/wradlib/wradlib-notebooks.

You can write notebooks with pure markdown and a little boilerplate code.

The notebooks can already be rendered via pytest. For usage within the docs, it would need to be copied over or it should be taken care of within conf.py. Also for CI this will be taken care of.

For ReadTheDocs we will use https://github.com/wradlib/wradlib-notebooks as the repo which will hold the rendered versions (which are precompiled in GH Actions). I'll get this running in the next days. I'll also add a proper documentation of this new workflow.


kmuehlbauer:
OK, then I'll keep a copy of this branch, where we can pull the already converted examples from. Then I'm removing everything from this PR and just keep the examples/notebooks with __init__.py, conftest.py and the stripped down overview.md. After merge everyone can base of this and add notebooks as needed/wanted.

kmuehlbauer:
This also changes docstrings in many source files, pointing to (now unavailable) notebooks. Before another release we need to fix these. 

I'm going to merge this now, so that others can add notebooks, see #712.

PR #736, opened 2025-12-03
Improve contributing experience

egouden:
* clarify contributing guide
    * fix requirements
    * add conda environment
    * catch missing HDF4 error
    * requires netcdf for cfradial1 example


kmuehlbauer:
@egouden I like it! But it might be not so easy to keep this documentation current.

kmuehlbauer (approved):

kmuehlbauer:
I was about to release anyway. Please improve, if you see fit.

PR #758, opened 2026-03-19
Added attrs for Radolan RE secondary variable (hail flag)

guidara:
Closes #753.

This PR adds attrs to the `secondary` ancillary variable of Radolan RE.

It adds a new method `_get_ancillary_attrs` of class `_radolan_file` that could be later expanded to add attrs to other ancillary variables and other radolan products.

### DIFF
--- a/wradlib/io/radolan.py
+++ b/wradlib/io/radolan.py
@@ -1365,7 +1365,9 @@ class _radolan_file:
 
         if self._ancillary:
             for a in self._ancillary:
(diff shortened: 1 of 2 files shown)


kmuehlbauer (changes requested):
@guidara Thanks! This looks good. Just one minor request. After reolving, this can go in.

kmuehlbauer (approved):
Thanks!

PR #768, opened 2026-05-23
FIX: minor issues related to contributing

egouden:


requirements_devel.txt
```diff
@@ -7,6 +7,7 @@ pytest
 pytest-cov
 pytest-doctestplus
 pytest-sugar
+pytest-testmon
```
kmuehlbauer: Is this referenced in the docs? Just one or two sentences on usage would be great.

kmuehlbauer (approved):
One suggestion and please also add entry to release notes, otherwise LGTM.

PR #772, opened 2026-06-02
ENH: compositing notebook reflectivity plots set to 60 dBZH max

egouden:
This is to be consistent with other DBZH plots.


kmuehlbauer (approved):
Thanks @egouden for spotting this.

PR #787, opened 2026-07-02
FIX: propagate nans in z_to_r_enhanced

egouden:
This is a quick fix for propagating nans instead of zeros in this conversion function.

The behavior of this rather complex function with missing data might need to be further investigated.


kmuehlbauer (approved):
Thanks @egouden, this change looks sensible to me.

PR #790, opened 2026-07-26
ENH: silent zero divide warning in filter_window_distance

egouden:
In region with all nans, a divide by zero occurs, propagating nans as expected. A warning was printed. It becomes silent after the change.


kmuehlbauer:
Thanks @egouden!

